PR #25, opened 2021-12-13
DBFS with multiple connections between sub grids

### LINKED ISSUE #22: Adapt DBFS communication protocol
...  to account for multiple connections between same `GridAgent`s.

### LINKED ISSUE #129: Correctly finish grid simulation in inferior grids
In case there are multiple connections between two `GridAgents` multiple `FinishGridSimulation` triggers are sent, currently. However, more than one of those triggers will not be understood and lead to an error message, although nothing breaks.

Therefore, send this trigger only once!

ckittl:
Resolves #22
Resolves #129 

Edit (SP): Please review #266 first!
Edit (SP): Test grid for `DBFSAlgorithmCenGridSpec` adapted to only have three inferior subgrids, one of which having two connections to the central subgrid


sebastian-peter (dismissed):
Some small things I noticed

sebastian-peter:
I guess #131 should be merged before this PR.

CHANGELOG.md
```diff
@@ -49,6 +48,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed some unreachable code [#167](https://github.com/ie3-institute/simona/issues/167)
 - Fix treatment of non-InitializeTrigger triggers in initialization within SimScheduler [#237](https://github.com/ie3-institute/simona/issues/237)
 - Fix breaking SIMONA caused by introducing temperature dependant load profiles in PSDM [#255](https://github.com/ie3-institute/simona/issues/255)
+- Fix power exchange between grids [#22](https://github.com/ie3-institute/simona/issues/22)
```
t-ober, line 51:
```suggestion
- Fix handling of multiple connections between subgrids [#22](https://github.com/ie3-institute/simona/issues/22)
```

src/test/scala/edu/ie3/simona/agent/grid/DBFSAlgorithmCenGridSpec.scala
```diff
@@ -8,18 +8,15 @@ package edu.ie3.simona.agent.grid
```
t-ober, line 8: Add test for multiple connection mechanisms that checks if 0 power is sent for the node that is not a slack node

src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
```diff
@@ -278,61 +279,108 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
 
     // / after power flow calc for this sweepNo
     case Event(
-          RequestGridPowerMessage(_, requestedNodeUuid),
-          PowerFlowDoneData(gridAgentBaseData, powerFlowResult)
+          RequestGridPowerMessage(_, requestedNodeUuids),
```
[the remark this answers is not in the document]
sebastian-peter: I added some documentation to the `PowerMessage`s that were missing it up until now

src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
```diff
@@ -278,61 +279,108 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
...
+          powerFlowResult match {
+            case validNewtonRaphsonPFResult: ValidNewtonRaphsonPFResult =>
+              val exchangePowers = requestedNodeUuids
+                .map { nodeUuid =>
+                  /* Figure out the node index for each requested node */
+                  nodeUuid -> gridAgentBaseData.gridEnv.gridModel.nodeUuidToIndexMap
+                    .get(nodeUuid)
+                    .flatMap { nodeIndex =>
+                      /* Find matching node result */
+                      validNewtonRaphsonPFResult.nodeData.find(stateData =>
```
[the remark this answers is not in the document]
sebastian-peter: It's a nice idea, we'd just have to realize in [powerflow](https://github.com/ie3-institute/powerflow), as that's where the `PowerFlowResult` classes reside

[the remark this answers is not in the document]
sebastian-peter: Created an issue here: https://github.com/ie3-institute/powerflow/issues/65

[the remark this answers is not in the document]
sebastian-peter: I now did a very similar thing with `SweepValueStore` though... :)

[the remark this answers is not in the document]
sebastian-peter: Ok regarding `SweepValueStore`, I reverted my changes because StateData needs to be in a certain order (Maps don't preserve order) and I can't figure out where the problem is currently. It'll be a problem for another day, I think in total DBFS can use some more revisions anyways.

src/test/scala/edu/ie3/simona/agent/grid/DBFSAlgorithmCenGridSpec.scala
sebastian-peter: I enhanced the test to also include a double connection to the superior subgrid... This also means that the result values changed now.
Furthermore, I refactored the test further so that it became simpler, yet again.

I also adapted slack voltage communication to only send one message per subgrid, in the same way this PR introduces one sending a single message for grid power communication per subgrid.

sebastian-peter: Ok, introducing the test for multiple slack nodes in powerflow computation revealed a shortcoming of the input data preparation, which is based on sets. Since sets have no determined order, at one point I had a test which worked on my PC and didn't in Jenkins. The solution here was to always pick the node with the smallest index (commit `b35c13ab`).

sebastian-peter:
I figured there's another important part of DBFS untested: Handling SystemParticipant's power/feed-in contributions -> #281

danielfeismann (changes requested):
Thanks for all the work done here (and also in review)! This looks good to me, I just have a minor question more for my own understanding.

src/test/scala/edu/ie3/simona/agent/grid/DBFSAlgorithmCenGridSpec.scala
```diff
@@ -165,765 +189,436 @@ class DBFSAlgorithmCenGridSpec
...
+        Seq(
+          ExchangePower(
+            supNodeA.getUuid,
+            Quantities.getQuantity(0, MEGAWATT),
+            Quantities.getQuantity(0, MEGAVAR)
+          ),
+          ExchangePower(
+            supNodeB.getUuid,
+            Quantities.getQuantity(0.160905770717798, MEGAVOLTAMPERE),
+            Quantities.getQuantity(-1.4535602349123878, MEGAVOLTAMPERE)
```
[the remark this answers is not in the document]
sebastian-peter: Yes, you're right. I found this annoying before, and now I found a way to return the correct types (with correct prefix). Check out commit https://github.com/ie3-institute/simona/pull/25/commits/fcee14cdbd9056a56daa5196d49a0945144805e9 if you're interested in my solution.

src/main/scala/edu/ie3/simona/model/grid/RefSystem.scala
```diff
@@ -121,10 +126,13 @@ final case class RefSystem private (
     * @param qInPu
     *   referenced active power value in p.u.
     * @return
-    *   unreferenced active power value in Var
+    *   unreferenced active power value in VAr
```
danielfeismann, line 129: I'm not 100% sure, but I think units that aren't dedicated to Persons are all written lowercase. So i would suggest var. (Just for the pedantic part in everyone of us :) )

src/test/scala/edu/ie3/util/quantities/QuantityUtilSpec.scala
```diff
@@ -186,4 +187,28 @@ class QuantityUtilSpec extends UnitSpec with TableDrivenPropertyChecks {
...
+        (MEGAVOLTAMPERE, WATT, MEGAWATT),
+        (VAR, WATT, WATT),
+        (KILOVAR, WATT, KILOWATT),
+        (MEGAVAR, WATT, MEGAWATT),
+        (VOLTAMPERE, VAR, VAR),
+        (KILOVOLTAMPERE, VAR, KILOVAR),
+        (MEGAVOLTAMPERE, VAR, MEGAVAR),
+        (WATT, VAR, VAR),
+        (KILOWATT, VAR, KILOVAR),
+        (MEGAWATT, VAR, MEGAVAR)
```
danielfeismann, line 206:
```suggestion
        (MEGAWATT, VAR, MEGAVAR),
        (VAR, VOLTAMPERE, VOLTAMPERE),
        (KILOVAR, VOLTAMPERE, KILOVOLTAMPERE),
        (MEGAVAR, VOLTAMPERE, MEGAVOLTAMPERE),
        (WATT, VOLTAMPERE, VOLTAMPERE),
        (KILOWATT, VOLTAMPERE, KILOVOLTAMPERE),
        (MEGAWATT, VOLTAMPERE, MEGAVOLTAMPERE)
```
Should we add also the remaining two cases, then the test covers every possibility?

danielfeismann (changes requested):

danielfeismann (commented):
Nice, I just have minor things which might be taken into account.

src/main/scala/edu/ie3/simona/model/grid/RefSystem.scala
sebastian-peter: I think you're right that ["the correct symbol is lower-case 'var', although the spellings 'Var' and 'VAr' are commonly seen"](https://en.wikipedia.org/wiki/Volt-ampere#Reactive).

However, we're using VAr in [PSU](https://github.com/ie3-institute/PowerSystemUtils/blob/24de7ea349f19eee0fd45395aeac5e8ddee11a6d/src/main/java/edu/ie3/util/quantities/PowerSystemUnits.java#L116), so that's why I'm using it here as well.

danielfeismann: Well, this doesn't change my opinion on this :) but I'm fine with both since it's clear what es meant here and thats the important thing. I just got on this since there is the proposed change from Var in VAr...

I checked DIN 1301 Teil 1 just for the sake of it: Unit name:  Var and Unit symbol: var.  I proposed this to [PSU](https://github.com/ie3-institute/PowerSystemUtils/pull/279)

src/test/scala/edu/ie3/simona/agent/grid/DBFSAlgorithmCenGridSpec.scala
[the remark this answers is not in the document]
danielfeismann: This is nice, thanks!

src/main/scala/edu/ie3/simona/model/grid/RefSystem.scala
sebastian-peter: Great! Regarding this PR, I reverted the naming to the correct unit name `Var`.

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -496,26 +496,21 @@ case object GridModel {
...
     // / nodes
-    val nodes: Set[NodeModel] =
-      subGridContainer.getRawGrid.getNodes.asScala
-        .collect { case nodeInput: NodeInput =>
-          NodeModel(nodeInput, startDate, endDate)
-        }
-        .to(collection.immutable.Set)
+    val nodes = subGridContainer.getRawGrid.getNodes.asScala.toSeq.map {
+      nodeInput => NodeModel(nodeInput, startDate, endDate)
+    }
```
t-ober, line 501: I think it might make sense to shortly comment on why this is transformed to a sequence as opposed to a set like for all the other grid elements

src/test/scala/edu/ie3/util/quantities/QuantityUtilSpec.scala
```diff
@@ -186,4 +187,34 @@ class QuantityUtilSpec extends UnitSpec with TableDrivenPropertyChecks {
       }
     }
   }
+
+  "Converting units to alternative units" should {
+    "succeed if units are compatible" in {
```
t-ober, line 192: Maybe add at least one test case where units are not compatible?

src/main/scala/edu/ie3/util/scala/quantities/QuantityUtil.scala
```diff
@@ -291,4 +292,18 @@ object QuantityUtil {
...
+
+    /** Transform some power unit to given unit with the same prefix
+      * @param targetUnit
+      *   the target system unit
+      * @return
+      *   this unit converted to given
+      */
+    def toEquivalentIn(targetUnit: measure.Unit[Q]): measure.Unit[Q] =
+      targetUnit.transform(unit.getConverterTo(unit.getSystemUnit))
+  }
```
t-ober, line 308: I think it makes sense to transfer this class (as well as the RichQuantity) into the PowerSystemUtils as it might come in handy for other projects as well and we already have som Quantity related utils there.

t-ober (changes requested):
Thanks for your work 🙃

src/main/scala/edu/ie3/simona/agent/grid/GridAgentData.scala
```diff
@@ -383,30 +383,24 @@ object GridAgentData {
         receivedSlackValues: ReceivedSlackValues
```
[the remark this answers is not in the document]
sebastian-peter: Sounds good

src/test/scala/edu/ie3/util/quantities/QuantityUtilSpec.scala
sebastian-peter: I don't think that's possible, since the unit type parameter (`Q`) prevents incompatible types to be applied as parameters

src/main/scala/edu/ie3/util/scala/quantities/QuantityUtil.scala
t-ober: https://github.com/ie3-institute/PowerSystemUtils/pull/289

If you want to make a quick review we can add it while were at it

sebastian-peter: Ok, that was really quick. I removed `RichUnit` and `RichQuantity` here and replaced all calls with ones to the equivalents in PSU

t-ober (approved):

danielfeismann (approved):
Great!

PR #26, opened 2021-12-13
Separating free lots request from evcs power calculation

### LINKED ISSUE #19: Move parking lot request out of calculatePowerWithSecondaryDataAndGoToIdle
Currently, EvFreeLotsRequest are handled within calculatePowerWithSecondaryDataAndGoToIdle of EvcsAgentFundamentals.

Free lot requests should be handled inside its own method.

sebastian-peter:
Resolves #19

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
+++ b/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
@@ -48,6 +48,15 @@ final case class ValueStore[+D](
       .filter(entry => entry._1 <= requestedTick)
       .maxByOption(entry => entry._1)
 
+  /** Get the last known entry (with the highest tick)
+    *
+    * @return
+    *   An Option to the last entry
+    */
+  def last(): Option[(Long, D)] =
+    store
+      .maxByOption(entry => entry._1)
+
   /** Acquires the stored information within the specified tick window
     *
     * @param requestStart
--- a/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
+++ b/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
@@ -45,8 +45,8 @@ final case class ValueStore[+D](
     */
   def last(requestedTick: Long): Option[(Long, D)] =
     store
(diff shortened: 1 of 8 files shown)


src/main/scala/edu/ie3/simona/agent/ValueStore.scala
```diff
@@ -48,6 +48,15 @@ final case class ValueStore[+D](
...
       .maxByOption(entry => entry._1)
 
+  /** Get the last known entry (with the highest tick)
+    *
+    * @return
+    *   An Option to the last entry
+    */
+  def last(): Option[(Long, D)] =
+    store
+      .maxByOption(entry => entry._1)
```
ckittl, line 58: I'd rather like 
```suggestion
      .maxByOption(_._1)
```

or more expressive
```suggestion
      .maxByOption{
        case (tick, _) => tick
      }
```

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
```diff
@@ -904,7 +881,7 @@ class EvcsAgentModelCalculationSpec
           1
         )
       )
-      scheduler.expectMsg(CompletionMessage(5L))
+      scheduler.expectNoMessage()
```
ckittl, line 884: Is this check needed then? In line 848 you removed it. So I would opt for having it at both positions (if necessary and sensible) or delete it here as well.

ckittl (changes requested):
In general looks good to me! There are only two minor remarks.

However, we currently don't have any kind of Changelog and it would be sad, if we would loose all that kind of documentation. Maybe we can start a draft changelog, that is then consolidated, when there is the first release (cf. https://github.com/ie3-institute/simona/blob/ba6281214d8f2a7b7c1060cd6783cd03f822fc1f/CHANGELOG.md)?

sebastian-peter:
I have to close this one as well, as it's still linking to the old branch name

PR #66, opened 2022-01-10
Add a runner implementation

### LINKED ISSUE #65: Support for experiment runs
Currently, we cannot run multiple configuration files in one simulation run. However, it would be nice to sequentially process multiple configurations instead of just one per execution.

Such a multi-configuration support should:
 -be configurable (point to exact location with one config per line)
- create new logs in the corresponding config folder + disable console logging automatically and point to the log location to avoid a too clumsy console (e.g. "Started simulation XYZ. Full log can be found here: <PATH_TO_LOG>)
- Optionally (by config arg) aggregate results

ckittl:
Resolves #65


ckittl:
@johanneshiry Please revise, if the changes to logging still are necessary / sensible.

PR #71, opened 2022-01-11
Respect for slack voltage angle in distributed power flow calculation

### LINKED ISSUE #69: Respect for voltage angle in distributed power flow
Currently, when an actor receives the slack node voltage from it's superior grid agent, the angle of the voltage is not accounted for.

ckittl:
Resolves #69


src/main/scala/edu/ie3/simona/agent/grid/PowerFlowSupport.scala
```diff
@@ -77,93 +82,195 @@
...
+    * @return
+    */
+  private def nodeTypeTargetVoltageStateData(
+      nodeModel: NodeModel,
+      nodeIndex: Int,
+      receivedValuesStore: ReceivedValuesStore,
+      transformers2w: Set[TransformerModel],
+      transformers3w: Set[Transformer3wModel],
+      residualPower: Complex,
+      stateData: Array[StateData],
```
sebastian-peter, line 236: I'd leave out this parameter here, as it's pretty much only used when adding to the array (line 261) or returned as-is (line 272). Instead, I would change the return type of this method to `(NodeType, Complex, Option[StateData])`, so that the optional stateData is only added to the array (in `composeOperatingPoint`) if it's present.

(this then also goes towards fixing one of the two new Codacy issues (having more than 8 params))

src/main/scala/edu/ie3/simona/agent/grid/PowerFlowSupport.scala
```diff
@@ -77,93 +82,195 @@
       gridMainRefSystem: RefSystem,
       targetVoltageFromReceivedData: Boolean = true,
       ignoreTargetVoltage: Boolean = false
-  ): Array[PresetData] = {
-
-    val mainRefSystemPowerUnit = gridMainRefSystem.nominalPower.getUnit
+  ): (Array[PresetData], WithForcedStartVoltages) =
+    nodes.foldLeft((Array.empty[PresetData], Array.empty[StateData])) {
```
sebastian-peter, line 86: One could think about replacing this by a combination of `map`, `unzip` and `flatten`, which I believe would be a bit cleaner

sebastian-peter (dismissed):
Some minor stuff I noticed

danielfeismann (approved):
Looks fine to me, thank you!

danielfeismann (dismissed):
Looks good to me

danielfeismann (approved):
This looks fine for me. Thanks for take care of this one.

PR #75, opened 2022-01-11
Fix location of "vn_simona"

### LINKED ISSUE #72: Fix location of "vn_simona" test grid
The grid nodes are currently partly in Dortmund and Berlin

ckittl:
Resolves #72


sebastian-peter (approved):
Well all nodes now rest at a single point, but for a test grid I think that is fine.

johanneshiry:
> Well all nodes now rest at a single point, but for a test grid I think that is fine.

@sebastian-peter indeed the grid does not span such a wide area that the weather data would change, but it does if parts of the grid are in BER 😂

johanneshiry (approved):
LGTM - thx! :)

PR #91, opened 2022-01-12
Add config possibility for transformer control groups

### LINKED ISSUE #90: Add config for transformer control groups
Configuration has to comprise for each control group:
- List of controlled transformers (by uuid)
- List of measurement devices to account for (by uuid)
- Specification of control scheme (e.g. limit violation prevention vs. preventive control)
- Parameters for control scheme (thresholds, dead bands etc.)

ckittl:
Resolves #90


sebastian-peter:
#131 should be merged before this PR.

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
```diff
@@ -541,6 +549,50 @@ case object ConfigFailFast extends LazyLogging {
...
+          "A transformer control group cannot have no transformers assigned."
+        )
+      if (vMin < 0)
+        throw new InvalidConfigParameterException(
+          "The minimum permissible voltage magnitude of a transformer control group has to be positive."
+        )
+      if (vMax < 0)
+        throw new InvalidConfigParameterException(
+          "The maximum permissible voltage magnitude of a transformer control group has to be positive."
+        )
```
sensarmad, line 589: To discuss: Should we really allow any permissible voltage level that is positive? IMO, we should limit it to +- 10-20% of the 1.0 p.u. mark. I checked if a sophisticated logic is given in **GridModel.scala** and to my understanding, it does not exist. The second reviewer can double-check.

src/main/scala/edu/ie3/simona/model/control/TransformerControlGroup.scala
```diff
@@ -0,0 +1,75 @@
...
+import edu.ie3.powerflow.model.NodeData.StateData
+import edu.ie3.powerflow.model.PowerFlowResult.SuccessFullPowerFlowResult
+import edu.ie3.simona.model.control.TransformerControlGroup.RegulationCriterion
+import tech.units.indriya.ComparableQuantity
+
+import java.util.UUID
+import javax.measure.quantity.Dimensionless
+
+/** Business logic for a transformer control group. It's main purpose is to
+  * determine, if there is any regulation need and if yes, in what circumference
```
sensarmad, line 19: Circumference? -- suggested replacement with region or area. 
Or does it mean something entirely different -- e.g., type of regulation as suggested in the following parentheses?

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -592,6 +634,133 @@ case object GridModel {
...
+          None
+        } else if (negativeRequests.nonEmpty) {
+          /* There are only requests for lower voltages => decide for the lowest required voltage */
+          negativeRequests.minOption
+        } else {
+          /* There are only requests for higher voltages => decide for the highest required voltage */
+          positiveRequests.maxOption
+        }
+      }
+
```
sensarmad, line 760: **Double-check this section:** To my understanding, if vMax is set to 12 p.u. and there is a higher voltage request, there is a possibility for the voltage to be set at 12 p.u. (hypothetically speaking). Or did I get this wrong?

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
```diff
@@ -870,7 +871,76 @@ class ConfigFailFastSpec extends UnitSpec with ConfigTestData {
...
+            List.empty[String],
+            1.02,
+            0.98
+          )
 
+          intercept[InvalidConfigParameterException] {
+            ConfigFailFast invokePrivate checkTransformerControl(dut)
+          }.getMessage shouldBe "A transformer control group cannot have no transformers assigned."
+        }
+
```
sensarmad, line 905: This test has its list of strings swapped, I guess. Not a major issue because it should fail if either of the strings are empty.

sensarmad (commented):
Mostly minor comments from my side with one logical discussion request.

src/main/scala/edu/ie3/simona/model/control/TransformerControlGroup.scala
danielfeismann: IMO it should mean something like: to what extent a regulation is needed. I will change it to have a better understanding.

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
danielfeismann: Included lower and upper boundary of +-20%. Tests have also been adapted accordingly.

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
danielfeismann: @sensarmad: Not sure if I got this right. In case vMax is (hypothetically) set to 12 p.u. then only nodes with absolute voltage over 12 p.u. would activate the regulation. With new the boundaries vMax is now limited hard till 1.2 (and vMin till 0.8). So if vMax is now set to 1.1 p.u. and the actual voltage is for one node at 1.15 and for another at 1.12, then the regulation will identify just positive deviations (of 0.05 and 0.03 to vMax and will choose the maximum of these (here 0.05) as the regulation need.

sensarmad (approved):

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -82,6 +96,15 @@ case object GridModel {
...
 
+  /** Collection of grid-related control strategies
+    *
+    * @param transformerControlGroups
+    *   Transformer control groups
+    */
+  final case class GridControls(
+      transformerControlGroups: Set[ControlGroupModel]
+  )
+
```
sebastian-peter, line 107: What do we need this class for when we could just hand over `transformerControlGroups: Set[ControlGroupModel]` (e.g. in GridModel constructor) as well?

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -592,6 +634,133 @@ case object GridModel {
...
+    * @return
+    *   A set of relevant nodal uuids
+    */
+  private def determineNodeUuids(
+      measurementUnitInput: java.util.Set[MeasurementUnitInput],
+      measurementConfigs: List[String]
+  ): Set[UUID] = Set.from(
+    measurementUnitInput.asScala
+      .filter(input =>
+        measurementConfigs.contains(input.getUuid.toString) && input.getVMag
```
sebastian-peter, line 702: This is something that we maybe should test as well during config fail fast for GridAgent: Do the measurement units that are assigned for transformer control groups measure voltage magnitude? If this is not the case, the device is filtered out here and this can lead to very confusing results, I can imagine

src/main/scala/edu/ie3/simona/model/participant/SystemParticipant.scala
```diff
@@ -105,11 +105,7 @@ abstract class SystemParticipant[CD <: CalcRelevantData](
   def activeToReactivePowerFunc(
       nodalVoltage: ComparableQuantity[Dimensionless]
   ): ComparableQuantity[Power] => ComparableQuantity[Power] =
-    qControl.activeToReactivePowerFunc(
-      sRated.multiply(scalingFactor),
-      cosPhiRated,
-      nodalVoltage
-    )
+    qControl.activeToReactivePowerFunc(sRated, cosPhiRated, nodalVoltage)
```
sebastian-peter, line 108: Why is the scaling factor removed here? Looks like something that belongs to #82, but not necessarily this PR?

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
```diff
@@ -930,7 +931,78 @@ class ConfigFailFastSpec extends UnitSpec with ConfigTestData {
             )
           }.getMessage shouldBe "The weather data scheme 'this won't work' is not supported. Supported schemes:\n\ticon\n\tcosmo"
         }
+      }
+
+      "checking the transformer control groups" should {
```
sebastian-peter, line 936: Nicely coded, these tests :)

sebastian-peter (changes requested):
Nicely done, just some minor remarks

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
danielfeismann: This credit goes to @ckittl

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
danielfeismann: Good idea. I moved the FailFast of GridAgent into it's own class. The solution we found together changed that way, that there is a check if the transformer belongs to the subgrid before checking the measurements. Please have a look if it make sense. @sebastian-peter

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
danielfeismann: As discussed I thing this enables us later to include also other controls. But I'm open for discussion here.

src/main/scala/edu/ie3/simona/model/participant/SystemParticipant.scala
danielfeismann: Good catch, think also something got mixed here... have been rolled back now.

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+    val measurementUnits =
+      gridAgentInitData.subGridContainer.getRawGrid.getMeasurementUnits.asScala
+
+    val transformerUnits2W =
+      gridAgentInitData.subGridContainer.getRawGrid.getTransformer2Ws.asScala
+
+    val transformerUnits3W =
+      gridAgentInitData.subGridContainer.getRawGrid.getTransformer3Ws.asScala
+
+    if (maybeControlConfig.isDefined) {
```
sebastian-peter, line 48: This if-statement is unnecessary, since `maybeControlConfig.foreach` already only executes if there's a `Some()`

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+      maybeControlConfig.foreach(control =>
+        control.transformer.foreach(transformer =>
```
sebastian-peter, line 50: I would rename `transformer` to `controlGroup` here...

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+          transformer.transformers.map(UUID.fromString).foreach {
+            transformers =>
```
sebastian-peter, line 52: ... and then provide the `transformer` in singular here.

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+              // Check if transformer is part of subgrid of this GridAgent
+              val transformerUnit2W = transformerUnits2W
+                .find(element => element.getUuid == transformers)
```
sebastian-peter, line 55: If you apply a function to the variable right away, you can just replace it with a wildcard `_`. Same below two more times.

```suggestion
                .find(_.getUuid == transformer)
```

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+              val transformerUnit3W = transformerUnits3W
+                .find(element => element.getUuid == transformers)
+              if (transformerUnit2W.isDefined || transformerUnit3W.isDefined) {
```
sebastian-peter, line 58: Some ScalaDoc would be nice here, especially about why you're checking for the right transformers here

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+
+                transformer.measurements
+                  .map(UUID.fromString)
+                  .foreach { measurements =>
```
sebastian-peter, line 62: It's one `measurement` :)

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+                    val measurementUnit = measurementUnits
+                      .find(element => element.getUuid == measurements)
+                      .getOrElse(
+                        throw new GridAgentInitializationException(
```
sebastian-peter, line 66: Currently, once at least one transformer belongs to the sub grid, all measurement units of the control group have to be present in the sub grid. Is this how we want it? I'm not quite sure myself. I think if a control group can be used on more than one sub grid, necessarily all the measurement units are present in only one of the sub grids. Thus, all measurement units can't be present in every sub grid.

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
```diff
@@ -549,6 +554,56 @@ case object ConfigFailFast extends LazyLogging {
...
+    */
+  private def checkControlSchemes(control: Simona.Control): Unit = {
+    control.transformer.foreach(checkTransformerControl)
+  }
+
+  /** Check the suitability of transformer control group definition.
+    *
+    * One important check cannot be performed at this place, as input data is
+    * not available, yet: Do the measurements belong to a region, that can be
+    * influenced by the transformer?
```
sebastian-peter, line 570: That's part of what happens in `GridAgentFailFast`, right? You could link it here with a short hint and `[[edu.ie3.simona.agent.grid.GridAgentFailFast]]`

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
```diff
@@ -930,7 +931,78 @@ class ConfigFailFastSpec extends UnitSpec with ConfigTestData {
...
+            0.98,
+            1.02
+          )
+
+          intercept[InvalidConfigParameterException] {
+            ConfigFailFast invokePrivate checkTransformerControl(dut)
+          }.getMessage shouldBe s"The minimum permissible voltage magnitude of a transformer control group (${dut.toString}) must be smaller than the maximum permissible voltage magnitude."
+        }
+
+        "throw Exception when vMin is lower then -21% of nominal Voltage" in {
```
sebastian-peter, line 979:
```suggestion
        "throw Exception if vMin is lower than -20% of nominal Voltage" in {
```

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
```diff
@@ -930,7 +931,78 @@ class ConfigFailFastSpec extends UnitSpec with ConfigTestData {
...
+            0.79
+          )
+
+          intercept[InvalidConfigParameterException] {
+            ConfigFailFast invokePrivate checkTransformerControl(dut)
+          }.getMessage shouldBe s"A control group (${dut.toString}) which control boundaries exceed the limit of +- 20% of nominal voltage! This may be caused " +
+            "by invalid parametrization of one control groups where vMin is lower than the lower boundary (0.8 of nominal Voltage)!"
+        }
+
+        "throw Exception when vMax is higher then +21% of nominal Voltage" in {
```
sebastian-peter, line 993:
```suggestion
        "throw Exception if vMax is higher than +20% of nominal Voltage" in {
```

src/test/scala/edu/ie3/simona/model/grid/GridSpec.scala
```diff
@@ -7,20 +7,25 @@
 package edu.ie3.simona.model.grid
 
 import java.util.UUID
-
 import breeze.linalg.DenseMatrix
 import breeze.math.Complex
 import breeze.numerics.abs
 import edu.ie3.datamodel.exceptions.InvalidGridException
+import edu.ie3.datamodel.models.input.MeasurementUnitInput
+import edu.ie3.datamodel.models.voltagelevels.GermanVoltageLevelUtils
```
sebastian-peter, line 15: Unnecessary imports, also below

sebastian-peter (changes requested):
I really like the validation of control groups that you implemented here. I have just some minor points, most relating to syntax.

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -592,6 +634,133 @@ case object GridModel {
...
+    /* Determine the voltage regulation criterion for each of the available nodes */
+    val nodeUuidToRegulationCriterion = nodeUuids.map { uuid =>
+      uuid -> { (complexVoltage: Complex) =>
+        {
+          val vMag = complexVoltage.abs
+          if (vMag > vMax)
+            Some(vMax - vMag)
+          else if (vMag < vMin)
+            Some(vMin - vMag)
+          else
```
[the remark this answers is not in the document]
danielfeismann: I don't think so. This determines the need for regulation. Since the else case apply when the voltage is inbetween the limits there is no need to regulate it. Or did I got you wrong?

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+import edu.ie3.simona.agent.grid.GridAgentData.GridAgentInitData
+import edu.ie3.simona.config.SimonaConfig
+import edu.ie3.simona.exceptions.agent.GridAgentInitializationException
+
+import java.util.UUID
+import scala.jdk.CollectionConverters._
+
+case object GridAgentFailFast {
+
+  def failFast(
```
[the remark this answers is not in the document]
danielfeismann: Please have a look

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
danielfeismann: Please have a look

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
danielfeismann: Yes, at least partly - is it possible to measure the influence. I've to thing about if there is a corner case where the measurement belongs to same subgrid as the transformer but can not be influenced by it for some reason.

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
```diff
@@ -0,0 +1,81 @@
...
+      gridAgentInitData.superiorGridGates.isEmpty && gridAgentInitData.inferiorGridGates.isEmpty
+    )
+      throw new GridAgentInitializationException(
+        s"${gridAgentInitData.subGridContainer.getGridName} has neither superior nor inferior grids! This can either " +
+          s"be cause by wrong subnetGate information or invalid parametrization of the simulation!"
+      )
+
+    /** Check if there exits voltage measurements for transformerControlGroups
+      */
+
```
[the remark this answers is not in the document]
danielfeismann: Please have a look

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
danielfeismann: Yes thats my understanding as well. We might should change this when there will be controlGroups in future regulating at HV/MV transformer and measurements checking the influence at LV. But I'm not sure how realistic it will be that this will happen. Another case: 3W transformer with measurement in both subgrids. So far I would life with this, but let's discuss.

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
sebastian-peter: I'd just plainly assume that measurements can always be influenced by the tranformers in that sub grid. Even if there are corner cases where that is not true, in my opinion it's ok to ignore those here. It's also something that I think I wouldn't require to be tested in a failFast context, as this goes beyond simple plausibility checking.

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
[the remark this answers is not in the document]
sebastian-peter: Oh yes, I was too fast here

src/main/scala/edu/ie3/simona/agent/grid/GridAgentFailFast.scala
sebastian-peter: I'm fine either way, I think we should just make it clear to the user if measurement units and transformers _have to_ sit in the same sub grid. Which brings me to the point of documentation... Maybe we should at least provide some very basic documentation in readthedocs here.

danielfeismann: Control groups are know part of documentation (basic version, might be extended later)

sebastian-peter:
I compared this PR with #92 and there doesn't seem to be that much difference. One is that here, at some points (e.g. `TransformerControlGroup.scala`) we do not use `squants` yet, but use the old quantities. 

My suggestion would be that we save us some work, close this PR and just deal with #92, since that PR seems to be more "ready" anyway.

danielfeismann:
@sebastian-peter We can do so. However, I think Chris split it into two PRs to have them smaller. Therefor I would prefer to keep them separated.

danielfeismann:
Closed in favor of #92

danielfeismann:
Closed in favor of https://github.com/ie3-institute/simona/pull/92

PR #109, opened 2022-01-25
Run log filter threshold from xml

### LINKED ISSUE #108: Run log filter threshold from xml
Take over the filters from defined standard file appender from logback xml-config, if applicable. Else log with "INFO"-Level

ckittl:
Resolves #108


ckittl:
Restart of rolling file appender does not work too great, yet. If required, I would have to look into this in detail.

@johanneshiry I assume this is the intended behavior? Or did you intend the log file in output directory only have "INFO"-Level and above? In that case I'd need to adapt the implementation.

sebastian-peter (dismissed):
LGTM

johanneshiry:
> Restart of rolling file appender does not work too great, yet. If required, I would have to look into this in detail.
> 
> 
> 
> @johanneshiry I assume this is the intended behavior? Or did you intend the log file in output directory only have "INFO"-Level and above? In that case I'd need to adapt the implementation.

The main intention is to **always** have a log file with the INFO-level in the output directory. If you want to have additional logs with more detailed log levels you can just adapt additional appenders in the xml-config. Therefore I would prefer not changing this behavior, but leave it to the user in the xml config to configure additional logger with e.g. debug level or below. 
Background: I always want at least **one** non-cluttered info log in the output dir. If you want debug log it can be **additionally** configured in the xml config file.

johanneshiry (changes requested):
See comment regarding your question. If required we can discuss this further. Pls let me know if you want to do so.

t-ober:
`
        rootLogger.iteratorForAppenders().asScala.foreach {
          case rf: RollingFileAppender[_] =>
            rf.getTriggeringPolicy.start()
            rf.start()
          case appender => appender.start()
        }
`
in LofbackConfiguration#default fixes the main issue.

CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed ChpData to ChpRelevantData [#494](https://github.com/ie3-institute/simona/issues/494)
 
 ### Fixed
+- Location of `vn_simona` test grid (was partially in Berlin and Dortmund)
```
danielfeismann, line 69: Does this apply to this PR?

danielfeismann (changes requested):
Looks good to me, just one very minor question.

CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Take log file event filters from `logback.xml` when defining the run log appender
```
danielfeismann, line 70: Issue No.

danielfeismann (approved):
LGTM

PR #112, opened 2022-01-25
Fix custom thread allocation in GridAgent

### LINKED ISSUE #111: Fix invalid thread allocation in GridAgent
The current implementation of the GridAgent uses its own implementation of an `ExecutionContext`.  However, this is a) not needed and b) leads to a huge amount of allocated Threads when an extensive grid is simulated on a large computational machine. Therefore, either the custom ExecutionContext should be replace by the dispatcher context OR the (somehow unrequired) Future should be removed from this actor as well.

johanneshiry:
Resolves #111

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
+++ b/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
@@ -52,7 +52,7 @@ import java.time.{Duration, ZonedDateTime}
 import java.util.UUID
 import javax.measure.Quantity
 import javax.measure.quantity.{Dimensionless, ElectricPotential}
-import scala.concurrent.Future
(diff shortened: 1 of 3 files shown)


sebastian-peter (dismissed):
LGTM. Since you completely removed parallel initialization here, do you think it is not worth the effort?

ckittl (dismissed):

ckittl:
!test

ckittl (approved):

johanneshiry:
> LGTM. Since you completely removed parallel initialization here, do you think it is not worth the effort?

@sebastian-peter 
Not really no. In particular bc we're not really using it correctly and do sth that is not recommended: a blocking operation in an actor. Therefore we should evaluate the need to process this asynchronously when we switch to akka typed. But for now this should be fine.

sebastian-peter (approved):

PR #121, opened 2022-01-27
Simplifying PrimaryServiceProxy with new changes from PSDM snapshot

sebastian-peter:


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - Improving code readability in EvcsAgent by moving FreeLotsRequest to separate methods
+- Using snapshot version of PSDM
```
ckittl, line 10:
```suggestion
- BREAKING: Using snapshot version of PSDM
```

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceProxy.scala
```diff
@@ -412,7 +348,7 @@ case class PrimaryServiceProxy(
     * @return
     *   The [[ActorRef]] to the spun off actor
     */
-  private def classToWorkerRef[V <: Value](
+  private[primary] def classToWorkerRef[V <: Value](
```
ckittl, line 351: What is the reason for package privacy?

ckittl (changes requested):
"Should work" is a reason to refuse by itself. ^^

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceProxy.scala
sebastian-peter: Because it's now directly mocked in PrimaryServiceProxySpec:
https://github.com/ie3-institute/simona/blob/a18bac454afff36661fdd9f50b4cc95a556555e5/src/test/scala/edu/ie3/simona/service/primary/PrimaryServiceProxySpec.scala#L459-L464

sebastian-peter: One could also make it `protected` here, I was just adapting it from the now deleted `private[primary] def columnSchemeToActor()`

ckittl: Hm... Okay... Well... Not too nice, what I introduced back then. However, fully omitting it, would require some refactoring of the primary service proxy, so I would be okay with kinda leaving it as it is, atm.

If possible, I'd go with protected here and for `initializeWorker`, as it is more restrictive. And, you might do me a favor and set `handleCoveredModel` to private.

sebastian-peter: `classToWorkerRef` is now `protected`.

`initializeWorker` and `handleCoveredModel` cannot be made `protected` here since it is called directly in `PrimaryServiceProxySpec` (unless you'd want to introduce private method testers here).

ckittl: I'd rather go with private method testers than having the scope too loose.

sebastian-peter: Ok, I tried both.

We cannot make `handleCoveredModel` private since `PrimaryServiceProxy` is subclassed in `PrimaryServiceProxySpec.scala:616`. Private method tester uses `getClass.getDeclaredMethods`, which means that only methods at the lowest level (here it is only `initializeWorker`) can be called. This means that we cannot call `handleCoveredModel` using private method tester.

For `initializeWorker`, there's a very similar problem to this, as we have a sublcass in `PrimaryServiceProxySpec.scala:459`.

So in short, I don't see any easy solution for making these methods private or protected.

sebastian-peter: Hold on, found a solution just when I finished typing. Let me know what you think about it.

johanneshiry (approved):
Thanks for your work. LGTM from the code side. 

Regarding the sonarqube issues pls do the following before merging: 
- close the two open code smells in sonarqube as "won't fix" with the comment you added in the code 

Regarding the change in the behavior:
- I'm fine with not exploding anymore but instead logging an error when a specific source cannot be created. However, there @ckittl should decide if this is fine so far, as in the old implementation a failure in the `modelClassToActor` lead to a shutdown. Now it doesn't anymore, but only logs the error. I don't know if this is mission critical, therefore @ckittl needs to decide.

sebastian-peter:
> close the two open code smells in sonarqube as "won't fix" with the comment you added in the code

Fixed the scapegoat/sonarqube warnings (which have to be suppressed in code btw, since the code smell was external to sonarqube)

> in the old implementation a failure in the `modelClassToActor` lead to a shutdown. Now it doesn't anymore, but only logs the error.

Don't really know what you mean here, maybe you could explain further.

Maybe you mean the former `columnSchemeToActor`? The possible failure there is not applicable anymore, as `ColumnScheme` now requires a value type to be provided. So, the only way that actor creation could formerly fail here, which is being provided with an unknown column scheme, does not apply anymore.

johanneshiry:
> > close the two open code smells in sonarqube as "won't fix" with the comment you added in the code
> 
> Fixed the scapegoat/sonarqube warnings (which have to be suppressed in code btw, since the code smell was external to sonarqube)

Perfekt thank you! :) 
 
> > in the old implementation a failure in the `modelClassToActor` lead to a shutdown. Now it doesn't anymore, but only logs the error.
> 
> Don't really know what you mean here, maybe you could explain further.
> 
> Maybe you mean the former `columnSchemeToActor`? The possible failure there is not applicable anymore, as `ColumnScheme` now requires a value type to be provided. So, the only way that actor creation could formerly fail here, which is being provided with an unknown column scheme, does not apply anymore.

Thanks for the clarification. Didn't know this, as I'm not that deep into the ColumnScheme in PSDM. I'm fine with merging then. If you want to you may check beforehand why CI is currently failing and fix this as well. After a fast look I think the reason might be a change in the test report dir (see https://simona.ie3.e-technik.tu-dortmund.de/ci/job/ie3-institute/job/simona/job/dev/70/console) which needs to be adapted in the Jenkinsfile.

sebastian-peter:
> If you want to you may check beforehand why CI is currently failing and fix this as well.

After restarting two times, it succeeded... Could be a Jenkins issue?

src/test/scala/edu/ie3/simona/service/primary/PrimaryServiceProxySpec.scala
```diff
@@ -681,17 +642,32 @@ class PrimaryServiceProxySpec
...
+          override protected def initializeWorker(
               columnScheme: ColumnScheme,
               timeSeriesUuid: UUID,
               simulationStart: ZonedDateTime,
               primaryConfig: PrimaryConfig,
               mappingSource: TimeSeriesMappingSource
           ): Try[ActorRef] = Success(probe.ref)
+
+          // needs to be overwritten as to make it available to the private method tester
+          @SuppressWarnings(Array("NoOpOverride"))
```
johanneshiry, line 654: 🎖️

johanneshiry (approved):

johanneshiry:
Thanks @sebastian-peter for the explanation and adaptions. Just approved it. We still need a go from @ckittl though :)

ckittl (approved):

PR #127, opened 2022-02-03
Re-organizing test resources into packages

### LINKED ISSUE #105: Re-organizing test resources to packages
Test data can be moved to the respective packages that it is needed in. This way, it is more obvious to which test the data belongs and vice versa. Also, `getClass.getResource()` works well with this, as only the path relative to the own package path needs to be stated.

For example, the contents of `it-data/primaryService` can be moved into some directory in `edu/ie3/simona/service/primary`. The full path then does not need to be stated when using `getClass.getResource()`.

As discussed with @t-ober, and @johanneshiry @ckittl you are welcome to give your opinion about this as well.

sebastian-peter:
Resolves #105


ckittl (approved):

johanneshiry (approved):
LGTM. Thank you!

PR #131, opened 2022-02-08
Fix config auto-generation

ckittl:
Fixes #130 
Fixes #148

PR #150, opened 2022-02-22
Consolidate csv parameterization

### LINKED ISSUE #149: Consolidate csv parameterization in config
Currently, abstract classes and inheritance is not properly used when defining csv parameters in config. Improve this and also add information about if the csv files are given in hierarchic directory structure.

ckittl:
Take care of #131 first, but then fixes #149


src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * © 2021. TU Dortmund University,
+ * © 2022. TU Dortmund University,
```
danielfeismann, line 2: I'm not familiar with copyrights but often I see something like: 2021 - 2022 
What to you think?

ckittl: Hmm... I'm not sure, if this is really relevant. The `gradle spotlessApply` takes care of the generation. Thereby, I guess we even cannot get the information anymore, when which file has be initiated...

sebastian-peter: It's because `SimonaConfig` keeps getting regenerated by tscfg. After generation, it doesn't have the typical ie3 header, so after applying spotless, a new header is used with the current year

sebastian-peter:
One thing to remember after merging this PR is to replace all `folderPath` in simona configs with `directoryPath`. For the included sample configs this has already happened, but not for configs external to this project.

danielfeismann (approved):
Looks good, thanks a lot!

t-ober (approved):
👍🏼

danielfeismann (approved):
Looks good

PR #155, opened 2022-02-25
Change transformer calculation

### LINKED ISSUE #99: Change transformer calculation
As of PowerSystemDataModel > 3.0 the no load susceptance of transformers (both two and three winding) will come negative. The automatic negation here should be adapted.

danielfeismann:
removed multiplier (-1) to comply with changes in PowerSystemDataModel >3.0

resolves #99


danielfeismann:
To reviewers: Unsure if changes in ThreeWindingTestData are necessary (Test says not, but logically it needs to be changed there as well)

johanneshiry (dismissed):
The changes LGTM, thanks for your work. 
However, did you also take into account the test input data (e.g. the vn_simona grid csv files). I'm wondering why the tests doesn't break here, because I would assume that at least the integration test _should_ fail? 🤔 
If this is not the case I'll approve it asap :) 

Regarding your question for 3 winding transformers maybe @ckittl can shed some light on this.

danielfeismann:
> The changes LGTM, thanks for your work. However, did you also take into account the test input data (e.g. the vn_simona grid csv files). I'm wondering why the tests doesn't break here, because I would assume that at least the integration test _should_ fail? 🤔 If this is not the case I'll approve it asap :)
> 
> Regarding your question for 3 winding transformers maybe @ckittl can shed some light on this.

Thanks for your review. Also for remember me about the test input data. I checked them, they include just the reactance X not the susceptance B so this change doesn't effect the test input data.

danielfeismann:
> To reviewers: Unsure if changes in ThreeWindingTestData are necessary (Test says not, but logically it needs to be changed there as well)

ThreeWindingTestData => #163

sebastian-peter (changes requested):
There are still some files that in my eye need adaptation:

input/samples/vn_simona/fullGrid/transformer_2_w_type_input.csv
input/samples/vn_146_lv_small/fullGrid/transformer_2_w_type_input.csv

ckittl (approved):
Thanks a lot for your changes!

To all of us: Please be aware, that after this change, currently available grid models have to be revised thoroughly before using! This especially applies to the SimBench grids, as there the changes are not yet considered: ie3-institute/simBench2psdm#81

t-ober (approved):

PR #171, opened 2022-03-10
Fix unreachable code

### LINKED ISSUE #167: Fix unreachable code
https://github.com/ie3-institute/simona/blob/1d5dcb7e72fac0eb0296e2dc4145ff3ed55806f0/src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala#L304-L307

Introduced with #77. Statements have to be switched. Also, a test for this is required.

https://simona.ie3.e-technik.tu-dortmund.de/sonar/project/issues?id=edu.ie3%3Asimona&issues=AX5EWrTicE05oj4IaDLM&open=AX5EWrTicE05oj4IaDLM

danielfeismann:
resolve #167

### DIFF
--- a/src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
+++ b/src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
@@ -6,7 +6,8 @@
 
 package edu.ie3.simona.event.listener
 
-import akka.actor.{ActorRef, FSM, PoisonPill, Props, Stash}
+import akka.actor.{ActorRef, FSM, Props, Stash, Status}
+import akka.pattern.pipe
 import akka.stream.Materializer
 import edu.ie3.datamodel.io.processor.result.ResultEntityProcessor
 import edu.ie3.datamodel.models.result.ResultEntity
@@ -23,6 +24,7 @@ import edu.ie3.simona.event.listener.ResultEventListener.{
   BaseData,
   Init,
   ResultEventListenerData,
+  SinkResponse,
   Transformer3wKey,
   UninitializedData
 }
@@ -38,12 +40,7 @@ import edu.ie3.simona.io.result.{
   ResultSinkType
 }
 import edu.ie3.simona.logging.SimonaFSMActorLogging
-import edu.ie3.simona.sim.SimonaSim
-import edu.ie3.simona.sim.SimonaSim.{
-  ServiceInitComplete,
-  ServiceInitFailed,
-  ServiceInitResponse
-}
+import edu.ie3.simona.sim.SimonaSim.ServiceInitComplete
 import edu.ie3.simona.util.ResultFileHierarchy
 
 import java.util.concurrent.TimeUnit
@@ -63,6 +60,10 @@ object ResultEventListener extends Transformer3wResultSupport {
   private final case object UninitializedData extends ResultEventListenerData
 
   private final case object Init
+
+  private final case class SinkResponse(
+      response: Map[Class[_], ResultEntitySink]
+  )
 
   /** [[ResultEventListener]] base data containing all information the listener
     * needs
@@ -294,10 +295,6 @@ class ResultEventListener(
       stash()
       stay()
 
-    case Event(baseData: BaseData, UninitializedData) =>
-      unstashAll()
-      goto(Idle) using baseData
-
     case Event(Init, _) =>
       Future
         .sequence(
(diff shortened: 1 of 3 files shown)


sebastian-peter:
While reviewing, I found that we have the same issue of non-wrapped exceptions thrown just a couple lines above:
https://github.com/ie3-institute/simona/blob/6125fa9726b540a6ddf54ab76621d2ab6d915d1e/src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala#L116-L122

I implemented a solution here, let me know what you think: https://github.com/ie3-institute/simona/compare/df/%23167-fix-unreachable-code...sp/%23167-ResultEventListener-suggestion 
If you think these changes are helpful, feel free to just fast-forward this branch to mine.

sebastian-peter:
That being said, I also found another weak spot of this implementation, revealed by the new test.

Since initialization is started within `preStart()`, there is a chance that initialization crashes before the creator of this listener even has the chance to `watch()` its child. This means there's a chance that we don't even notice the initialization crashing.

https://github.com/ie3-institute/simona/blob/6125fa9726b540a6ddf54ab76621d2ab6d915d1e/src/test/scala/edu/ie3/simona/event/listener/ResultEventListenerSpec.scala#L152-L153

src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
```diff
@@ -287,10 +295,6 @@ class ResultEventListener(
       stash()
       stay()
 
-    case Event(baseData: BaseData, UninitializedData) =>
-      unstashAll()
-      goto(Idle) using baseData
-
     case Event(Init, _) =>
```
johanneshiry, line 298: What's the reasoning behind this deletion?

src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
```diff
@@ -299,18 +303,20 @@ class ResultEventListener(
...
+    case Event(SinkResponse(classToSink), _) =>
+      // Sink Initialization succeeded
+      log.debug("Initialization complete!")
+      supervisor ! ServiceInitComplete
+
+      unstashAll()
+      goto(Idle) using BaseData(classToSink)
+
+    case Event(Status.Failure(ex), _) =>
+      throw new InitializationException("Unable to setup SimonaSim.", ex)
```
johanneshiry, line 319: Maybe an additional logging statement with the exception could be slightly more verbose here?

johanneshiry (approved):
LGTM thank you! :-)

Please consider taking into account my questions and comments if you want to. But these are only of minor concern.

src/test/scala/edu/ie3/simona/event/listener/ResultEventListenerSpec.scala
```diff
@@ -141,6 +137,22 @@ class ResultEventListenerSpec
         assert(outputFile.exists)
         assert(outputFile.isFile)
       }
+
+      "check if actor dies when it should die" in {
+        val fileHierarchy = resultFileHierarchy(1, ".ttt")
```
t-ober, line 142:
```suggestion
        val fileHierarchy = resultFileHierarchy(5, ".ttt")
```

t-ober: Since the specific runid is used for a `ResultFileHierarchy` at line 121 the test can fail since the directory that should be created already exists. As the direcetory that is created uses the runid passed to the `resultFileHierarchy` for naming using a different id solves the problem

t-ober (changes requested):
Looks good - this pipe thing is cool. Please have a look at the ResultFileHierarchy naming issue though

src/test/scala/edu/ie3/simona/event/listener/ResultEventListenerSpec.scala
sebastian-peter: Already fixed this in the branch I mentioned [here](https://github.com/ie3-institute/simona/pull/171#issuecomment-1065615202)

src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
sebastian-peter: This has become obsolete due to the result of the sink initialization now being piped to itself

src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
danielfeismann: Good point, please have a look again

ckittl: There also is a logging statement in the defined supervision strategy. We might have a look, if this is not double the work, then.

danielfeismann: seems it would log twice

```
      case ex: Exception =>
        self ! EmergencyShutdownInitiated
        logger.error(
          "The simulation's guardian received an uncaught exception. - {}: \"{}\" - Start emergency shutdown.",
          ex.getClass.getSimpleName,
          ex.getMessage
        )
```

t-ober (approved):

ckittl (approved):

PR #254, opened 2022-06-06
Create heatpump agent

### LINKED ISSUE #253: Implement agent for heat pumps
Model is ready yet, but there is no specific agent for it.

As the `HpModel` consumes electric and produces thermal heat at the same time, please take special care of the scaling factor. This is not yet considered in the existing other agents.

### Open sub-tasks
- [ ] Loading of thermal models and assigning to `GridAgentInitData`

Edit TO:
- [ ] add thermal storage as flexibility resource

ckittl:
Resolves #253
Requires PSDM >4.0.0


src/main/scala/edu/ie3/util/scala/quantities/QuantityUtil.scala
```diff
@@ -108,15 +94,16 @@ object QuantityUtil {
...
+        averagingUnit(
+          integrate(
+            values,
+            windowStart,
+            windowEnd,
+            integrationQuantityClass,
+            integrationUnit
+          ).value.doubleValue
+            / (windowEnd - windowStart)
+        )
```
danielfeismann, line 106: Unsure about this. Let's double check this.

src/main/scala/edu/ie3/util/scala/quantities/QuantityUtil.scala
```diff
@@ -193,15 +180,14 @@ object QuantityUtil {
               (tick, value)
             ) =>
           /* Calculate the partial integral over the last know value since it's occurrence and the instance when the newest value comes in */
-          val duration = Quantities.getQuantity(tick - lastTick, Units.SECOND)
+          val duration = (tick - lastTick).toDouble
           val partialIntegral =
-            lastValue.multiply(duration).asType(integrationQuantityClass)
-          val updatedIntegral = currentIntegral.add(partialIntegral)
+            integrationUnit(lastValue.value.doubleValue * duration)
```
danielfeismann, line 185: Unsure about this. Let's double check this.

src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
```diff
@@ -0,0 +1,386 @@
...
+    val thermalHouseInput = {
+      inputModel match {
+        case ParticipantStateData.SimpleInputContainer(_) =>
+          None
+        case WithHeatInputContainer(_, thermalGrid) =>
+          /* Build the thermal model */
+          thermalGrid.houses().stream().findFirst().asScala
+      }
+    }.getOrElse {
+      new ThermalHouseInput(
```
danielfeismann, line 310: Do we want to create some random thermal bus if the correct one can't be found? From my point of view an exception would be better here. Or?

danielfeismann (changes requested):
Beside one minor thing this looks good to me. A second reviewer would be great since I also worked on this changes here...

src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
danielfeismann: After discussing this shortly we opt for an exception in this case.

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgentFundamentals.scala
```diff
@@ -1725,4 +1760,83 @@ case object ParticipantAgentFundamentals {
...
+    * @param windowStart
+    *   First, included tick of the time window
+    * @param windowEnd
+    *   Last, included tick of the time window
+    * @param activeToReactivePowerFuncOpt
+    *   An Option on a function, that transfers the active into reactive power
+    * @return
+    *   The averaged apparent power
+    */
+  def averageApparentPowerAndHeat(
```
sebastian-peter, line 1777: Here, we could re-use `averageApparentPower` instead of duplicating all this code.

src/main/scala/edu/ie3/simona/io/grid/CsvGridSource.scala
```diff
@@ -0,0 +1,115 @@
...
+  ThermalBusInput,
+  ThermalHouseInput,
+  ThermalStorageInput
+}
+
+import java.nio.file.Path
+import scala.jdk.CollectionConverters._
+
+object CsvGridSource {
+  def readGrid(
```
sebastian-peter, line 23: This method is not used, we implemented it in PSDM

sebastian-peter (changes requested):
There's some more things one could discuss, but these are all topics for another day.

sebastian-peter (approved):
Thanks for taking care of my comments!

PR #266, opened 2022-06-21
Applying a lot of code improvements to DBFS algo & surrounding classes

### LINKED ISSUE #265: Improve code quality in and around DBFSAlgorithm

sebastian-peter:
Resolves #265

While reviewing #25, I noticed that a lot of changes were related to code style improvements rather than the fixes advertised fixes of #25. The code improvements are authored by both @ckittl and @sebastian-peter.

These changes should be merged before #25, after which #25 will be a lot smaller and easier to understand.

Hint for reviewers: This PR is supposed to only change syntax, not the meaning of the code. This is also why I have not touched tests here.

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
+++ b/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
@@ -45,13 +45,11 @@ import edu.ie3.simona.ontology.messages.VoltageMessage.{
 import edu.ie3.simona.ontology.trigger.Trigger._
 import edu.ie3.simona.util.TickUtil._
 import edu.ie3.util.quantities.PowerSystemUnits._
-import edu.ie3.util.quantities.{PowerSystemUnits, QuantityUtil}
 import tech.units.indriya.quantity.Quantities
 
 import java.time.{Duration, ZonedDateTime}
 import java.util.UUID
-import javax.measure.Quantity
-import javax.measure.quantity.{Dimensionless, ElectricPotential}
+import javax.measure.quantity.ElectricPotential
 import scala.concurrent.{ExecutionContext, Future}
 
 /** Trait that is normally mixed into every [[GridAgent]] to enable distributed
@@ -119,6 +117,7 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
       // we just received either all provided slack voltage values or all provided power values
       val updatedGridAgentBaseData: GridAgentBaseData = receivedValues match {
         case receivedPowers: ReceivedPowerValues =>
+          /* Can be a message from an asset or a message from an inferior grid */
           gridAgentBaseData.updateWithReceivedPowerValues(receivedPowers)
         case receivedSlacks: ReceivedSlackValues =>
           gridAgentBaseData.updateWithReceivedSlackVoltages(receivedSlacks)
@@ -154,7 +153,6 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
           allValuesReceived,
           updatedGridAgentBaseData
         )
-
       }
 
     // if we receive a request for slack voltages from our inferior grids we want to answer it
@@ -171,10 +169,7 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
 
       // we either have voltages ready calculated (not the first sweep) or we don't have them here
       // -> return calculated value or target voltage as physical value
-      val (slackE, slackF): (
-          Quantity[ElectricPotential],
-          Quantity[ElectricPotential]
-      ) =
+      val (slackE, slackF) =
         (gridAgentBaseData.sweepValueStores.get(currentSweepNo) match {
           case Some(result) =>
             Some(result, currentSweepNo)
@@ -226,7 +221,7 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
                 uuid == nodeUuid && isSlack
               }
               .map(_.vTarget)
-              .getOrElse(Quantities.getQuantity(1d, PowerSystemUnits.PU))
+              .getOrElse(Quantities.getQuantity(1d, PU))
           val vSlack = vTarget
             .multiply(
               gridAgentBaseData.gridEnv.gridModel.mainRefSystem.nominalVoltage
@@ -368,12 +363,11 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
           gridAgentBaseData: GridAgentBaseData
         ) =>
       // inform my child grids about the end of this grid simulation
-      gridAgentBaseData.inferiorGridGates.foreach(inferiorGridGate => {
-        gridAgentBaseData.gridEnv
-          .subnetGateToActorRef(inferiorGridGate) ! FinishGridSimulationTrigger(
-          currentTick
-        )
-      })
+      gridAgentBaseData.inferiorGridGates
+        .map { inferiorGridGate =>
+          gridAgentBaseData.gridEnv.subnetGateToActorRef(inferiorGridGate)
+        }
+        .foreach(_ ! FinishGridSimulationTrigger(currentTick))
 
       // inform every system participant about the end of this grid simulation
       gridAgentBaseData.gridEnv.nodeToAssetAgents.foreach { case (_, actors) =>
@@ -1005,7 +999,8 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
     *   a map contains a mapping between nodes and the [[ActorRef]] s located @
     *   those nodes
     * @param refSystem
-    *   the reference system of the [[GridModel]] of this [[GridAgent]]
+    *   the reference system of the [[edu.ie3.simona.model.grid.GridModel]] of
+    *   this [[GridAgent]]
     * @param askTimeout
     *   a timeout for the request
     * @return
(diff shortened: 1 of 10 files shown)


src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
```diff
@@ -542,15 +542,52 @@ private[grid] trait GridResultsSupport {
...
-          -signum(iPu.real)
-        ) * 180.0,
-        PowerSystemUnits.DEGREE_GEOM
-      )
+      angle(iPu)
     )
 
+  /** Correct the offset of an angle dependent on the direction. If the
+    * direction is negative, 180° are added
+    */
```
t-ober, line 550: Okay so I had to draw that stuff to see why what we are doing here makes sense so I would extend the documentation a little bit.

To me what we are doing seems to be to adjust an angle so instead of getting the relative angle we get its angle with respect to the real axis. This means that in case of a negative real value we need to offset it by + 180°

src/main/scala/edu/ie3/simona/agent/grid/GridAgentData.scala
```diff
@@ -219,63 +222,84 @@ object GridAgentData {
...
-            .updated(nodeUuid, receivedVector)
+      // update the values in the received map
+      val receivedVector = nodeToReceived
+        .getOrElse(
+          nodeUuid,
+          throw new RuntimeException(
+            s"NodeId $nodeUuid is not part of nodeToReceivedPowerValuesMap!"
+          )
+        )
+        /* Filter out the entry, that belongs to the given sender and is not defined, yet, in case no replacement shall
```
danielfeismann, line 293: Whenever there is time, just for my better understanding: Could you please give me some insights why this step is necessary?

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
```diff
@@ -97,22 +103,28 @@ case object ReceivedValuesStore {
...
+              subOrdinateToReceivedPower,
+              (inferiorSubGridRef, couplingNodeUuid)
+            ) =>
+          /* Check, if there is already something expected for the given coupling node and add reference to the subordinate
+           * grid agent */
+          val actorRefToMessage = subOrdinateToReceivedPower
+            .getOrElse(
+              couplingNodeUuid,
+              Vector.empty[(ActorRef, Option[ProvidePowerMessage])]
+            ) :+ (inferiorSubGridRef -> None)
```
t-ober, line 123: in this case I actually think the original implementation might be easier to follow and you spare yourself a vector conversion and mapping but this might be personal preference

src/main/scala/edu/ie3/simona/model/participant/SystemParticipant.scala
```diff
@@ -154,10 +154,10 @@ abstract class SystemParticipant[CD <: CalcRelevantData](
         KILOVOLTAMPERE
       )
 
-      if (apparentPower.isGreaterThan(sMax)) {
+      if (apparentPower.isGreaterThan(sMax.multiply(1.05))) {
         logger.warn(
           s"The var characteristics \'$qControl\' of model \'$id\' ($uuid) imposes an apparent " +
-            s"power (= $apparentPower) that exceeds " +
+            s"power (= $apparentPower) that exceeds 105 % of " +
```
t-ober, line 160: Is that what we want? There seems to be change of behavior between 100% and 105% of sMax.
Until 105% we just deal with having more apparent power than sMax and afterwards we limit it to 100% sMax. So if power actually increases from 105% to 106% we lose power and from 105% to 100%.

Why don't we stick with 100% in the first place? I think exceeding it makes sense for some systems like pvs (where I think we deal with it different anyways) but maybe not for bio mass plants or something like this.

t-ober (changes requested):
Thanks a lot for extracting this from the other PR and adding the code improvements 🙂

danielfeismann (approved):
Thanks for the improvements!

src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
```diff
@@ -1108,29 +1092,30 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
...
+              (inferiorGridAgentRef ? RequestGridPowerMessage(
                 currentSweepNo,
-                inferiorGridGate.getSuperiorNode.getUuid
+                inferiorGridGateNodes
               )).map {
                 case provideGridPowerMessage: ProvideGridPowerMessage =>
-                  (inferiorGridAgent, Option(provideGridPowerMessage))
+                  (inferiorGridAgentRef, provideGridPowerMessage)
                 case FailedPowerFlow =>
-                  (inferiorGridAgent, Some(FailedPowerFlow))
```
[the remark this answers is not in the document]
sebastian-peter: Yup, I think it was just used to match `edu.ie3.simona.agent.grid.ReceivedValuesStore#NodeToReceivedPower`, although it doesn't actually make sense to send optional power messages

src/main/scala/edu/ie3/simona/agent/grid/GridAgentData.scala
sebastian-peter: It's actually a good question. We're filtering out the exact value in the vector that we want to replace with the received data here. This is actually only necessary because we use a vector here, a map would take care of this automatically, since each key is mapped to exactly one value. So it'd be a good idea to replace the vector with a map, and this portion of the code can be removed. We can look through this together if you like :)

src/main/scala/edu/ie3/simona/model/participant/SystemParticipant.scala
sebastian-peter: Don't know where excatly this comes from. I guess the threshold was increased so that slight exceedings of 100% would not trigger the log statement and the reduction of reactive power.

I'd suggest that I reset this to the way it's handled in dev and we create another issue for this, where we continue this discussion?

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
```diff
@@ -542,15 +542,52 @@ private[grid] trait GridResultsSupport {
...
+    * @return
+    *   The angle of the complex value
+    */
+  private def angle(cplx: Complex): ComparableQuantity[Angle] = cplx match {
+    case _ if cplx.abs == 0 =>
+      /* The complex value has no magnitude, therefore define the angle to zero */
+      Quantities.getQuantity(0d, PowerSystemUnits.DEGREE_GEOM)
+    case Complex(real, imag) =>
+      /* Calculate the angle between real and imaginary part */
+      val baseAngle = atan(imag / real).toDegrees
```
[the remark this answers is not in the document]
sebastian-peter: If `imag = 0.0` and `real = 0.0`, [then the division results in `NaN`](https://stackoverflow.com/questions/51801054/what-decides-nan-and-infinity-in-java-division-operations). 
With `if cplx.abs == 0`, this case is already excluded above, so you're right, `NaN` cannot occure here.

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
sebastian-peter: The original implementation did not account for `real == 0` and the corresponding special cases. I did adapt the code a little bit though, let me know if it is easier to understand now.

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
sebastian-peter: I also just had to quickly sketch this out :D 
I refactored the code a little bit and added commentary, lmk what you think

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
```diff
@@ -542,51 +541,52 @@ private[grid] trait GridResultsSupport {
...
           imag
         )
-      } else {
-        /* Correct the angle for full orientation in the complex plane */
-        offSetCorrection(
+      case Complex(real, imag) =>
+        /* Both real and imaginary part are != 0. This means atan can be used to
+        compute an angle between -90 and 90 degrees. To compute the actual angle
+        (between -180 and 180 degrees), 180 degrees have to be added in case that
+        the real part is negative. */
```
t-ober, line 571: I think this doesn't quite nail it but feel free to correct me if im wrong.

What we seem to want is the absolute angle of the imaginary part calculated with respect to the positive real axis which relates to a positive real part.
So atan actually computes the direct angle between the imaginary part (Gegenkathete) and the real part (Ankathete). In case the imaginary part is negative we calculate the relative angle with respect to the negative real axis which would mean we have to "turn" the angle by 180°

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
```diff
@@ -47,7 +47,7 @@ final case class ReceivedValuesStore private (
 object ReceivedValuesStore {
 
   type NodeToReceivedPower =
-    Map[UUID, Vector[(ActorRef, Option[PowerResponseMessage])]]
+    Map[UUID, Map[ActorRef, Option[PowerResponseMessage]]]
```
t-ober, line 50: Since we changed to use a map now doesn't that mean the whole boolean replace part does no longer work? If so is that safe? Also if so we might want to remove the functionality.

t-ober (changes requested):
Okay I'm very sorry for all the annoying questions and concerns. But all stuff related to the DBFS seem pretty dangerous to me so I just want to calm myself by really understanding all of the changes before I approve changes.

src/main/scala/edu/ie3/simona/agent/grid/GridAgentData.scala
```diff
@@ -257,19 +257,19 @@ object GridAgentData {
         nodeToReceived: NodeToReceivedPower,
         senderRef: ActorRef,
         replace: Boolean
-    ): Map[UUID, Vector[(ActorRef, Option[PowerResponseMessage])]] = {
+    ): NodeToReceivedPower = {
```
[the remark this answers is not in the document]
sebastian-peter: That's true

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
sebastian-peter: Yes, that's what's happening...
(besides the point that you were probably trying to say "In case the **real** part is negative [...]" in your last sentence)

How would you enhance the commentary here?

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
sebastian-peter: Well the `replace` boolean is now only used in `getNodeUuidForSender()` (which was `uuid()` before), and it only functions as some sort of safeguard, where an exception is thrown if your expectation is not met. Meaning the if you expect to not have received data before but you actually *did*, the uuid wouldn't be returned and you would throw in `updateNodalReceivedPower`. 

So yes, I think `replace` doesn't serve a functionality that is really required here, and I think it's up to personal preference whether one would want to have it removed.

t-ober: Okay I get it now - thanks for clarification 🙂

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
```diff
@@ -100,27 +100,29 @@ object ReceivedValuesStore {
...
+        (uuid, actorRefs.map(actorRef => actorRef -> None).toMap)
     }
 
-    /* Add everything, that I expect from my sub ordinate grid agents. Build distinct pairs of sending actor reference
-     * and target node */
-    inferiorSubGridGateToActorRef.toVector
+    /* Add everything, that I expect from my sub ordinate grid agents.
+     * Build distinct pairs of sending actor reference and target node.
+     * Convert to sequence first, since the map operation conflates
+     * key/value pairs with the same key */
```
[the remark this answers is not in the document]
sebastian-peter: Great question. 

There can be multiple connections between two subgrids, which up until #25 requires multiple messages to be sent - which wasn't really handled though.
The thing here is that the DBFS tests that we currently have are not able to perform here, since TestProbes are not used and just once test actor acts as a stand-in for multiple GridAgents of inferior grids. I think this is why we need `toSeq` here.

The whole situation is really not satifying, but for this issue here I'd suggest waiting for #25 to enable multiple connections between subgrids. I want to improve the tests there to include said multiple connections and then check again if this code can be improved still (I think so). 
Implementing this in this PR would be double the work, since with #25 the behavior changes again.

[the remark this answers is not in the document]
sebastian-peter: Pls check how I refactored this in #25: https://github.com/ie3-institute/simona/blob/2ad69c43703ebae16bd47987476a3dc668c1ea31/src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala#L107-L127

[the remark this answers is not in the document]
sebastian-peter: Also, the reason why this works is because yesterday in #25, I adapted the test to use multiple inferior-GA-TestProbes

https://github.com/ie3-institute/simona/blob/2ad69c43703ebae16bd47987476a3dc668c1ea31/src/test/scala/edu/ie3/simona/agent/grid/DBFSAlgorithmCenGridSpec.scala

src/main/scala/edu/ie3/simona/model/participant/SystemParticipant.scala
sebastian-peter: New issues:  #275

Enhance the issue description, if you see the need.

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
danielfeismann: Maybe like this?
```suggestion
        /* Both real and imaginary part are != 0. This means atan can be used to
        compute an angle between -90 and 90 degrees. To compute the actual angle
        (between -180 and 180 degrees), 180 degrees have to be added in case that
        the real part is negative. */
```

t-ober: I like the proposal

src/main/scala/edu/ie3/simona/agent/grid/ReceivedValuesStore.scala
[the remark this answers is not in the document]
t-ober: Alright thanks for the feedback I agree lets then handle this in #25

t-ober (approved):
Thanks ! 
I like daniels request for the commentary but nevertheless I think we can finally conclude this one 🙂 

Edit: The CI seems to have found some issues we should take care of.

src/main/scala/edu/ie3/simona/agent/grid/GridResultsSupport.scala
sebastian-peter: Yes, thanks @danielfeismann!

t-ober (approved):

danielfeismann (approved):
Thanks for all the work done here and also to Thomas for your detailed review!

PR #312, opened 2022-08-15
Renaming sub-package directories.

### LINKED ISSUE #141: Rename sub-package directories in test resources
As decided in https://github.com/ie3-institute/PowerSystemDataModel/pull/525, rename sub-package directories so that they are prefixed with `_`

staudtMarius:
Resolves #141


sebastian-peter (changes requested):
Thanks, looks all good. Please just add a line to `CHANGELOG.md` according to the exisiting scheme, then this is good to go

sebastian-peter (approved):
Looks good, thanks!

Issue #301, opened 2022-08-02
Enhance `AUTHORS.md` with missing authors


sebastian-peter:
Currently, @danielfeismann is missing here as well as many past and current Scientific Assistants

PR #317, opened 2022-08-25
Updated authors #301

vickybung1:


sebastian-peter (dismissed):
Looks good! There's only one thing missing before we can merge this: you'd need to execute `gradle spotlessApply` one time and commit and push again, since the format is off a tiny bit.

sebastian-peter (approved):
All good now, thanks!

PR #325, opened 2022-08-30
Removing config of vn_146_lv_small

### LINKED ISSUE #290: Broken config of vn_146_lv_small

vickybung1:
Resolves #290


sebastian-peter:
Ok upon inspection of the results that we get from this grid I realized there are no system participants connected whatsoever, so there is no power flow happening and we get no output that one could use for anything.
Given this fact @t-ober and I decided that it would probably be best to just remove this sample grid from the SIMONA code base alltogether, which will reduce overall maintenance needs. So could you please just delete the whole vn_146 grid directory and commit? Thanks!

src/main/scala/edu/ie3/simona/main/RunSimonaStandalone.scala
```diff
@@ -17,6 +17,7 @@ import edu.ie3.simona.ontology.messages.SchedulerMessage.{
   SimulationFailureMessage,
   SimulationSuccessfulMessage
 }
+
```
sebastian-peter, line 20: Also please revert this :)

sebastian-peter (approved):
Thanks, all good now

PR #351, opened 2022-09-26
Fixing `ResultEventListener` exiting too early with high volumes of results

### LINKED ISSUE #350: Fix missed messages of `ResultEventListener`
With large amounts of messages sent, `ResultEventListener` takes too long to write out results and/or is terminated too early.

sebastian-peter:
Resolves #350

Main changes:
- Using BufferedWriter instead of Futures, which do not make sense when writing to a single file
- Implementing gentler stopping of `ResultEventListener`: SimonaSim now asks it to stop, and the `ResultEventListener` is using a ReceiveTimeout to wait until (hopefully) all results have been received.


src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
```diff
@@ -117,14 +114,19 @@ object ResultEventListener extends Transformer3wResultSupport {
               )
               .flatMap { fileName =>
                 if (fileName.endsWith(".csv") || fileName.endsWith(".csv.gz")) {
-                  ResultEntityCsvSink(
-                    fileName.replace(".gz", ""),
-                    new ResultEntityProcessor(resultClass),
-                    fileName.endsWith(".gz")
-                  ).map((resultClass, _))
+                  Future {
```
t-ober, line 117: Does this work as expected when we have multiple `ResultEventListener`? We hand back an Iterable of Futures which we then map to `SinkResponse`. And as it seems to me we don't wait for every sink to be finished but are going to idle unstashing all messages as soon as we get the first `SinkResponse`

t-ober (changes requested):
Looks great! The code got much less complicated.

src/main/scala/edu/ie3/simona/io/result/ResultEntityCsvSink.scala
```diff
@@ -197,28 +149,31 @@ object ResultEntityCsvSink {
...
-      delimiter: String = ",",
-      bufferSize: Int = 50
-  )(implicit materializer: Materializer): Future[ResultEntityCsvSink] = {
+      delimiter: String = ","
+  ): ResultEntityCsvSink = {
+
+    val file = new File(outfileName)
+    val existedBefore = file.exists()
+
+    val writer = new BufferedWriter(new FileWriter(file, existedBefore), 32768)
```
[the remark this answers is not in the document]
sebastian-peter: It's `2^15`, we could also adjust the size. Haven't spent that much time thinking about what could be a reasonable buffer size. Choosing a rather large size is fine since the buffer is supposed to be flushed in the end anyways.

src/main/scala/edu/ie3/simona/event/listener/ResultEventListener.scala
sebastian-peter: No, the iterable of Futures is still handed the same way. Difference is: `ResultEntityCsvSink.apply` returns the sink directly, not a future to one. Thus, we need to wrap in a future here.

danielfeismann (approved):
Looks good to me, thanks for this improvement!

danielfeismann (changes requested):
Just a minor thing: Changelog :)

sebastian-peter:
!test

danielfeismann (approved):
LGTM

PR #362, opened 2022-09-30
Re-implementing ``ResultEventListener`` in akka typed

### LINKED ISSUE #343: Re-implement `ResultEventListener` in akka typed
The transformed `RuntimeEventListener` could be used as an example.

Other useful resources:
- [Akka classic to typed guide](https://doc.akka.io/docs/akka/current/typed/from-classic.html)
- [Coexistence of classic and typed](https://doc.akka.io/docs/akka/current/typed/coexisting.html) (probably only useful in edge cases)
- The [OsmoGrid](https://github.com/ie3-institute/OSMoGrid) project, which has been rewritten in akka typed

staudtMarius:
Resolves #343 

Merge first:
- #351


sebastian-peter (changes requested):
@staudtMarius I made some minor changes to the syntax and simplified the partial 3w result handling. If you like, have a look and give me a review on my changes.

There's also one major comment I have regarding blocking during initialization, please have a look.

staudtMarius:
> @staudtMarius I made some minor changes to the syntax and simplified the partial 3w result handling. If you like, have a look and give me a review on my changes.
> 
> There's also one major comment I have regarding blocking during initialization, please have a look.

Looks good. I fixed the issue with the blocking behaviour.

src/main/scala/edu/ie3/simona/event/ResultEvent.scala
```diff
@@ -15,7 +15,7 @@ import edu.ie3.datamodel.models.result.connector.{
 import edu.ie3.datamodel.models.result.system.SystemParticipantResult
 import edu.ie3.simona.agent.grid.GridResultsSupport.PartialTransformer3wResult
 
-sealed trait ResultEvent extends Event
+trait ResultEvent extends Event
```
sebastian-peter, line 18: I think it's not that great to have all messages directly extend `ResultEvent`. 

One way to do it would to use [adapters](https://doc.akka.io/docs/akka/current/typed/interaction-patterns.html#adapted-response), but that has the drawback here that we have to send back the adapter's ActorRef to SimonaSim, and we want to keep the ResultEventListener as simple as possible.

So a better way to do this (I think) would be to have a trait `ResultMessage` and have `ResultEvent` extend that trait, as well as having ResultEventListener-specific messages (`SinkResponse` etc.) to extend `ResultMessage` as well.

sebastian-peter (changes requested):

sebastian-peter (approved):

PR #373, opened 2022-10-12
Replacement of end tick by current tick within the simulation scheduler

### LINKED ISSUE #159: Use correct tick when simulation ends
When the simulation ends, the current tick instead of the `endTick` should be used. This makes a difference if the simulation ends prematurely.

https://github.com/ie3-institute/simona/blob/17a915f279a5b7eecd46e42f87a968c0c2792946/src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala#L532

Finally, add a test for this.

vickybung1:
Resolves #159

### DIFF
--- a/src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
+++ b/src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
@@ -7,7 +7,6 @@
 package edu.ie3.simona.scheduler
 
 import akka.actor.ActorRef
-import breeze.linalg.min
 import edu.ie3.simona.event.RuntimeEvent.{
   CheckWindowPassed,
   Done,
@@ -294,12 +293,7 @@ trait SchedulerHelper extends SimonaActorLogging {
       awaitingResponseMap: CountingMap[Long],
       nowInTicks: Long
   ): Boolean =
-    awaitingResponseMap.minKeyOption match {
-      case Some(minKey) =>
-        nowInTicks - minKey <= 0L
-      case None =>
-        true // map empty, no completions awaited
-    }
+    awaitingResponseMap.isEmpty
 
   /** Checks if the provided state data and the current tick is eligible to
     * issue a [[CheckWindowPassed]] event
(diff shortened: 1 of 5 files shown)


vickybung1:
What is the error about?

src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
```diff
@@ -7,6 +7,7 @@
 package edu.ie3.simona.scheduler
 
 import akka.actor.ActorRef
+import breeze.linalg.min
```
sebastian-peter, line 10: Please just use `math.min` instead of introducing this import

src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
```diff
@@ -452,21 +446,31 @@ trait SchedulerHelper extends SimonaActorLogging {
...
+          totalSimDuration,
+          stateData.runtime.noOfFailedPF,
+          errorInSim
+        )
+      )
     } else {
       stateData.runtime.initSender ! SimulationSuccessfulMessage
+
+      /* notify listeners */
+      notifyListener(
```
sebastian-peter, line 466: Can't you use this in both cases?

src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
```diff
@@ -789,7 +792,7 @@ trait SchedulerHelper extends SimonaActorLogging {
     context.watch(actorToBeScheduled)
 
     // if the tick of this trigger is too much in the past, we cannot schedule it
-    if (stateData.time.nowInTicks - trigger.tick > parallelWindow) {
+    if (stateData.time.nowInTicks - trigger.tick > 0L) {
```
sebastian-peter, line 795: This can also be simplified

sebastian-peter (changes requested):
I like your changes, thanks for the work. I just have some minor questions and tips for simplifying the code. Let me know if I can help

src/main/scala/edu/ie3/simona/scheduler/SchedulerHelper.scala
sebastian-peter: I think since the code in 452-460 and in 464-472 are the same, you can just place it above the if-switch, like it was before.

Maybe also add a comment that the usage of `min` is necessary because the scheduler overshoots the target tick by 1 at the end of the simulation.

sebastian-peter (changes requested):
Only one more tiny issue, I think then we're good to go!

sebastian-peter (approved):
Looks good, thanks!

Issue #378, opened 2022-10-14
Restructure and reimplement `SimScheduler`


sebastian-peter:
Future concept shall allow the hierarchical stacking of schedulers by separating scheduling and time advancement functionality.

Also, the activation in phases could be controlled by the time advancement unit, activating separate schedulers.

PR #380, opened 2022-10-15
TMP: Introducing priority triggers to SimScheduler

### LINKED ISSUE #379: Introduce priority scheduling (TMP)
Introduce a second trigger queue for priority triggers. The completion of these triggers is awaited before regular triggers are sent.

This is needed for EmAgents and EvcsAgents (with ExtEvDataService) to work together.

This is only a temporary solution and the implementation should probably not be merged into `dev`. #378 will provide a much cleaner and more modular solution.

sebastian-peter:
Resolves #379 

This is only a temporary (and hacky) solution, which should be replaced by a solution for #378 at some point.
DO NOT MERGE


sebastian-peter:
Solved by #667

PR #319, opened 2022-08-26
Flexibility functionality for system participants

sebastian-peter:
Resolves #308 

Includes introduction of agent states, which need to be considered by agent models
(also includes applying scaling factor to active power in `PvModel` now)

PR #381, opened 2022-10-15
Heat pumps with flexibility

ckittl:
This is a combination of #319 and #316. Don't close before those changes have been approved!


sebastian-peter:
Now included in #714

PR #390, opened 2022-10-22
Add heatpump with flexibility

ckittl:


sebastian-peter:
Closing so that the pipeline might run again

PR #398, opened 2022-10-29
Speed up quantitites

### LINKED ISSUE #396: At most `Double` number precision in Quantities
Currently, a lot of Quantity numbers are converted to `BigDecimal` during operations like unit conversions or multiplications (cf. profiling of a SIMONA run with mobility sim). Since Quantities are constructed with `double` precision in PSDM, I don't think this is intended for our use case, especially because `BigDecimals` (which are based on `BigInteger`) are significantly slower than doubles during standard operations.

All of this is given narrowing is disabeld, which seems to affect execution speed as well.

_Originally posted by @sebastian-peter in https://github.com/ie3-institute/simona/discussions/292_

sebastian-peter:
Resolves #396


sebastian-peter:
This is probably not very useful anymore, as we mostly switched to squants

PR #400, opened 2022-10-30
Fix determineModelBaseStateData for RandomLoadModel

### LINKED ISSUE #399: Fix determineModelBaseStateData for RandomLoadModel
Usage of random load profiles does not deliver results since there is no case for RandomLoadModel in determineModelBaseStateData

### LINKED ISSUE #401: Fix first non random profiles per typeday of random load profile
The first day per typeday (weekday, saturday, sunday) of random load profile doesn't seem random.

danielfeismann:
resolves #399 
resolves #401

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/participant/load/random/RandomLoadModel.scala
+++ b/src/main/scala/edu/ie3/simona/model/participant/load/random/RandomLoadModel.scala
@@ -79,7 +79,7 @@ final case class RandomLoadModel(
       )
   }
 
-  private val randomFactory = RandomFactory.get(430431L)
+
   private val randomLoadParamStore = RandomLoadParamStore()
 
   type GevKey = (DayType.Value, Int)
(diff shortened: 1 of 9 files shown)


src/main/scala/edu/ie3/simona/model/participant/load/random/RandomLoadModel.scala
```diff
@@ -146,6 +145,7 @@ final case class RandomLoadModel(
       case Some(foundIt) => foundIt
       case None          =>
         /* Instantiate new gev distribution, put it to storage and return it */
+        val randomFactory = RandomFactory.get((Math.random() * 1000000).toInt)
```
t-ober, line 148: We can implement a 
`val randomFactory: RandomFactory = RandomFactory.get(42L)`
in the companion object so we don't need to set up individual factories for every load model

t-ober (changes requested):
Thanks 🙂

t-ober (approved):
💯

PR #424, opened 2022-12-15
Updated csv-files according to the new syntax

### LINKED ISSUE #216: `vn_simona` in SIMONA seems to uses deprecated csv file version
New csv geo json:
"{""type"":""Point"",""coordinates"":[7.411111,51.492528],""crs"":{""type"":""name"",""properties"":{""name"":""EPSG:4326""}}}"

Old csv geo json:
"{"type":"Point","coordinates":[7.4116482,51.4843281],"crs":{"type":"name","properties":{"name":"EPSG:0"}}}"

Reading files with the old format which is not specification XYZ compliant leads to problems when used in other grids.
This sould be updated. 

One could probably just read it in with the current PSDM implementation and persist it again to get rid of this problem.

jo-bao:
resolves #216


danielfeismann:
!test

input/samples/vn_simona/fullGrid/wec_type_input.csv
```diff
@@ -1,2 +1,2 @@
...
....0000000000,0.4204150540),(12.0000000000,0.3542412990),(13.0000000000,0.2884705910),(14.0000000000,0.2309657020),(15.0000000000,0.1877836700),(16.0000000000,0.1547289760),(17.0000000000,0.1289985520),(18.0000000000,0.1086711060),(19.0000000000,0.0923997500),(20.0000000000,0.0792212360),(21.0000000000,0.0684342820),(22.0000000000,0.0595200870),(23.0000000000,0.0520892490),(24.0000000000,0.0458456230),(25.0000000000,0.0405612730),(26.0000000000,0.0360588240),(27.0000000000,0.0321988460),(28.0000000000,0.0166182640),(29.0000000000,0.0103309760),(30.0000000000,0.0060915190),(31.0000000000,0.0033311770),(32.0000000000,0.0016416370),(33.0000000000,0.0007054230),(34.0000000000,0.0001966440),(35.0000000000,0.0000000000),(36.0000000000,0.0000000000),(37.0000000000,0.0000000000),(38.0000000000,0.0000000000),(39.0000000000,0.0000000000),(40.0000000000,0.0000000000),(41.0000000000,0.0000000000),(42.0000000000,0.0000000000),(43.0000000000,0.0000000000),(44.0000000000,0.0000000000),(45.0000000000,0.0000000000),(46.0000000000,0.0000000000),(47.0000000000,0.0000000000),(48.0000000000,0.0000000000),(49.0000000000,0.0000000000),(50.0000000000,0.0000000000)}",100.0,98.0,EnerconE-82,0.0,5281.02,2222.2
```
sebastian-peter, line 2: Do you know why the `0` decimal places are printed out like that?

jo-bao: To avoid rounding in other dimensions (e.g. (2.0000000000, **0.1159335160**)), I used 10 decimal places in the transformation (PSDM).

sebastian-peter: We now got rid of the decimal places parameter, which means that only the amount of digits that make sense for a number should be printed: https://github.com/ie3-institute/PowerSystemDataModel/pull/714
Would you mind trying with the changed PSDM?

sebastian-peter (changes requested):
Looks good to me. I just noticed that an addition to the changelog is missing, maybe you could provide that?

sebastian-peter (approved):
Looks good! I haven't checked all lines since they are in different orders now, but the files should have the same contents :)

PR #432, opened 2023-01-14
Changed simulation duration format

### LINKED ISSUE #429: Log total simulation duration of a SIMONA run
It would be nice to know how much time the simulation took. This could be logged in the simulation shutdown phase.

vickybung1:
Resolves #429

### DIFF
--- a/src/main/scala/edu/ie3/simona/io/runtime/RuntimeEventLogSink.scala
+++ b/src/main/scala/edu/ie3/simona/io/runtime/RuntimeEventLogSink.scala
@@ -77,19 +77,14 @@ final case class RuntimeEventLogSink(
     )
   }
 
-  private def convertDuration(duration: Double): String = {
-    val hh = ((duration / 1000 / 60) % 24).toInt
-    val mm = ((duration / 1000 / 60) % 60).toInt
-    val ss = roundAt(5)(duration / 1000) % 60
-    s"$hh h: $mm m: $ss s"
+  def convertDuration(duration: Long): String = {
+    val hh = duration / 1000 / 3600
+    val mm = (duration / 1000 / 60) % 60
+    val ss = (duration / 1000) % 60
+    s"${hh}h : ${mm}m : ${ss}s"
   }
 
-  private def roundAt(precision: Int)(number: Double): Double = {
-    val s = math pow (10, precision)
-    (math round number * s) / s
-  }
-
-  private def durationAndMemoryString(duration: Double) = {
+  private def durationAndMemoryString(duration: Long) = {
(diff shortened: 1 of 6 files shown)


src/main/scala/edu/ie3/simona/io/runtime/RuntimeEventLogSink.scala
```diff
@@ -76,8 +77,11 @@ final case class RuntimeEventLogSink(
     )
   }
 
-  private def roundDuration(duration: Double): Double = {
-    roundAt(5)(duration / 1000)
+  private def convertDuration(duration: Double): String = {
+    val hh = ((duration / 1000 / 60) % 24).toInt
```
sebastian-peter, line 81: Please leave out the `% 24`, as we want to see the full hour amount :) Also, I think another division by 60 is necessary to get to hours.

src/test/scala/edu/ie3/simona/event/listener/RuntimeEventListenerSpec.scala
```diff
@@ -102,8 +102,12 @@ class RuntimeEventListenerSpec
...
       }
 
-      def roundDuration(dur: Double): Double =
-        roundAt(5)(dur / 1000)
+      def convertDuration(duration: Double): String = {
+        val hh = ((duration / 1000 / 60) % 24).toInt
+        val mm = ((duration / 1000 / 60) % 60).toInt
+        val ss = roundAt(5)(duration / 1000) % 60
+        s"$hh h: $mm m: $ss s"
+      }
```
sebastian-peter, line 110: I'd advise against having the same function a second time, because that would need to be maintained as well. Maybe rather test against a set amound of milliseconds, so you'd test against something like `[...] Total runtime: 4 h: 19 m: 3 s`

src/main/scala/edu/ie3/simona/io/runtime/RuntimeEventLogSink.scala
```diff
@@ -76,8 +77,11 @@ final case class RuntimeEventLogSink(
...
+    val mm = ((duration / 1000 / 60) % 60).toInt
+    val ss = roundAt(5)(duration / 1000) % 60
+    s"$hh h: $mm m: $ss s"
```
sebastian-peter, line 84: Maybe a format like `4h : 19m : 3s` would be better than `4 h: 19 m: 3 s`

sebastian-peter (changes requested):
Some commentary on the `convertDuration` function

sebastian-peter (approved):
Looks fine!

PR #436, opened 2023-01-23
Adapting to changed data source in PSDM.

### LINKED ISSUE #435: Adapting to changed data sources in PDSM
The PR
- https://github.com/ie3-institute/PowerSystemDataModel/pull/712

will change how data sources return their data. Instead of returning optional that may be empty, The data source will now return the data directly if possible or throw an exception. Therefore simona needs to adapt to these changes.

staudtMarius:
Resolves #435

### DIFF
--- a/src/main/scala/edu/ie3/simona/io/result/ResultEntityCsvSource.scala
+++ b/src/main/scala/edu/ie3/simona/io/result/ResultEntityCsvSource.scala
@@ -8,7 +8,6 @@ package edu.ie3.simona.io.result
 
 import java.io.{File, FileInputStream}
 import java.util.zip.GZIPInputStream
-
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream.alpakka.csv.scaladsl.{CsvParsing, CsvToMap}
@@ -23,9 +22,15 @@ import akka.stream.scaladsl.{
 }
 import akka.stream.{FlowShape, Materializer}
 import com.typesafe.scalalogging.LazyLogging
+import edu.ie3.datamodel.exceptions.FactoryException
+import edu.ie3.datamodel.io.factory.FactoryData.MapWithRowIndex
 import edu.ie3.datamodel.io.factory.SimpleEntityData
 import edu.ie3.datamodel.io.factory.result.SystemParticipantResultFactory
 import edu.ie3.datamodel.models.result.ResultEntity
+import edu.ie3.datamodel.models.result.system.SystemParticipantResult
+import edu.ie3.datamodel.utils.StreamUtils
+import edu.ie3.datamodel.utils.StreamUtils.zipWithRowIndex
+import edu.ie3.datamodel.utils.options.Try
 import edu.ie3.simona.exceptions.{FileIOException, ProcessResultEventException}
 import org.apache.commons.io.FilenameUtils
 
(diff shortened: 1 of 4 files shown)


staudtMarius:
This needs 
- https://github.com/ie3-institute/PowerSystemDataModel/pull/712

to work.

staudtMarius:
!test

src/main/scala/edu/ie3/simona/io/result/ResultEntityCsvSource.scala
```diff
@@ -74,16 +75,18 @@ final case class ResultEntityCsvSource(
...
+
+    if (option.isSuccess) {
+      option.getData.get()
+    } else {
+      logger.error(
+        s"Unable to parse line in $filePath\nline: ${rowWithHeadline.values}\nWill ignore that line!"
+      )
+      throw new ProcessResultEventException(
+        s"Unable to parse line in $filePath\nline: ${rowWithHeadline.values}\nWill ignore that line!"
+      )
```
t-ober, line 89: Three things:

1. This class probably can be removed since it is not used anywhere. And if it would be needed we want that kind of functionality within the PSDM
2. Logging an error and throwing an exception seems unneccessary an exception is sufficient
3. If you get an exception from that try (which is not an option so it should not be called option) you would want to at least append the exception to the exception you are throwing otherwise the information is lost

t-ober: In that case you can double check if any of the code is used and if not just delete the file

src/main/scala/edu/ie3/simona/service/weather/WeatherSource.scala
```diff
@@ -111,7 +118,7 @@ trait WeatherSource {
 
     /* Go and get the nearest coordinates, that are known to the weather source */
     val nearestCoords = idCoordinateSource
-      .getNearestCoordinates(queryPoint, amountOfInterpolationCoords)
+      .getClosestCoordinates(queryPoint, amountOfInterpolationCoords, distance)
```
t-ober, line 121: Please add more information to the logging statement in case we don't find enough coordinates. Specifically mention the distance (and its unit) that has been used to find coordinates and that either there are no coordinates or it is out of bounds of the region for which we have weather data.

src/main/scala/edu/ie3/simona/io/result/ResultEntityCsvSource.scala
staudtMarius: Should I add this functionality to the PSDM?

src/main/scala/edu/ie3/simona/service/weather/SampleWeatherSource.scala
```diff
@@ -128,9 +133,25 @@ object SampleWeatherSource {
...
+        Vector.empty[CoordinateDistance].asJava
+    }
+
     override def getNearestCoordinates(
         coordinate: Point,
-        n: Int
+        i: Int
     ): util.List[CoordinateDistance] = {
       if (coordinate.getY.abs <= 90 && coordinate.getX.abs <= 180)
         Vector(
```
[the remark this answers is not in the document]
staudtMarius: Both methods in ``SampleWeatherSource`` are just returning the given coordinate. Also both methods have the same implementation in this class.

t-ober (dismissed):

PR #462, opened 2023-02-17
Check the grid validity with PSDM validation utils

### LINKED ISSUE #460: Check the grid for validity with the PSDM validation utils
We should check the grid we use for the simulation with the `ValidationUtils.check()` of the PSDM, and at least issue a warning in case the validation fails. Running simulation with a misconfigured grid model can lead to errors that are hard to track.

staudtMarius:
Resolves #460


src/main/scala/edu/ie3/simona/io/grid/GridProvider.scala
```diff
@@ -29,18 +32,20 @@ object GridProvider {
...
+            val jointGridContainer = CsvJointGridContainerSource.read(
+              simulationName,
+              params.csvSep,
+              params.directoryPath
+            )
+
+            Try.apply(ValidationUtils.check(jointGridContainer)) match {
+              case Failure(exception) =>
+                logger.warn(
+                  s"Error while initializing CsvGridSource! Cannot proceed without a valid GridSource!",
```
sebastian-peter, line 44: Please adapt the warning message. The error didn't happen during initialization, but during validation.

src/main/scala/edu/ie3/simona/io/grid/GridProvider.scala
```diff
@@ -29,18 +32,20 @@ object GridProvider {
...
-              .getOrElse(
-                throw new InitializationException(
-                  "Error while initializing CsvGridSource! Cannot proceed without a valid GridSource!"
+            val jointGridContainer = CsvJointGridContainerSource.read(
+              simulationName,
+              params.csvSep,
+              params.directoryPath
+            )
+
+            Try.apply(ValidationUtils.check(jointGridContainer)) match {
```
sebastian-peter, line 41:
```suggestion
            Try(ValidationUtils.check(jointGridContainer)) match {
```

src/main/scala/edu/ie3/simona/io/grid/GridProvider.scala
```diff
@@ -29,18 +32,20 @@ object GridProvider {
...
+              case Failure(exception) =>
+                logger.warn(
+                  s"Error while initializing CsvGridSource! Cannot proceed without a valid GridSource!",
+                  exception
```
sebastian-peter, line 45: I think it suffices to log the exception message here, otherwise it looks like too much of a serious error if we let the whole stacktrace be printed

```suggestion
                  s"Error while initializing CsvGridSource! Cannot proceed without a valid GridSource! \n\t{}",
                  exception.getMessage
```

src/main/scala/edu/ie3/simona/io/grid/GridProvider.scala
```diff
@@ -29,18 +32,20 @@ object GridProvider {
...
-                throw new InitializationException(
-                  "Error while initializing CsvGridSource! Cannot proceed without a valid GridSource!"
+            val jointGridContainer = CsvJointGridContainerSource.read(
+              simulationName,
+              params.csvSep,
+              params.directoryPath
+            )
+
+            Try.apply(ValidationUtils.check(jointGridContainer)) match {
+              case Failure(exception) =>
```
sebastian-peter, line 42: You should also care for the case that the validation actually succeeds (which currently doesn't happen that often, I guess). If there is no accounting for the success case, we'll fail then because it's not matched.

So I'd add something like this, maybe with a more elaborate debug message:

```scala
case _ =>
  log.debug("Validation succesful")
```

sebastian-peter (changes requested):
Looks good, just some hints that could improve the code

src/test/groovy/edu/ie3/simona/model/participant/PvModelIT.groovy
```diff
@@ -174,6 +172,7 @@ trait PvModelITHelper {
       HashMap modelToWeatherMap
       if (weatherMap.containsKey(time)) {
         modelToWeatherMap = weatherMap.get(time)
+        weatherMap.put(time, modelToWeatherMap)
```
sebastian-peter, line 175: Why is this needed? Seems like we're adding something back to the map, that we've retrieved from the map beforehand.

src/test/groovy/edu/ie3/simona/model/participant/PvModelIT.groovy
```diff
@@ -127,33 +126,32 @@ trait PvModelITHelper {
...
-          inputModel.getElevationAngle(),
+          QControl.apply(inputModel.qCharacteristics),
+          inputModel.sRated,
+          inputModel.cosPhiRated,
+          inputModel.node.geoPosition.y,
+          inputModel.node.geoPosition.x,
+          inputModel.albedo,
+          inputModel.etaConv,
+          inputModel.azimuth,
+          inputModel.elevationAngle,
```
sebastian-peter, line 154: Thanks for fixing these Codacy issues!

sebastian-peter (changes requested):

sebastian-peter (approved):
Looks good, I just slightly adapted the log messages and let's go

PR #476, opened 2023-03-01
more expressive primary service init failure

### LINKED ISSUE #475: Catch date time parsing exception to make error more expressive
It would be nice if the date time parsing exception that is thrown during primary data setup would be more expressive.

t-ober:
Resolves #475

### DIFF
--- a/src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
+++ b/src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
@@ -35,7 +35,9 @@ import edu.ie3.simona.service.{ServiceStateData, SimonaService}
 import edu.ie3.simona.util.TickUtil.{RichZonedDateTime, TickLong}
 import edu.ie3.util.scala.collection.immutable.SortedDistinctSeq
 
+import java.io.IOException
 import java.time.ZonedDateTime
+import java.time.format.DateTimeParseException
 import java.util.UUID
 import scala.jdk.CollectionConverters._
 import scala.jdk.OptionConverters.RichOptional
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -78,15 +80,30 @@ final case class PrimaryServiceWorker[V <: Value](
...
+              filePath,
+              valueClass,
+              factory
+            )
+          ) match {
+            case Failure(exc: DateTimeParseException) =>
+              throw new IOException(
+                s"Could not initiate the time series source. Parsing ${exc.getParsedString} failed with timepattern: $timePattern. The time pattern can be configured in the configuration.",
+                exc
+              )
```
sebastian-peter, line 98: If we want this kind of distinction, would'nt it rather make sense to have it in PSDM?
So, everything that can go wrong within *TimeSeriesSource is fully described by an exception that its constructor throws. Then, when we arrive at the level of PrimaryServiceWorker, we can then wrap the exception once more and say: "During creation of the timeseries source, an error occured." or something like that.

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -78,15 +80,30 @@ final case class PrimaryServiceWorker[V <: Value](
...
+            new CsvTimeSeriesSource(
+              csvSep,
+              directoryPath,
+              fileNamingStrategy,
+              timeSeriesUuid,
+              filePath,
+              valueClass,
+              factory
+            )
+          ) match {
```
sebastian-peter, line 93: This error matching would also be nice for SqlTimeSeriesSource as well, no? So maybe we can move this an abstraction level upwards, i.e. checking for `Failure` after *TimeSeriesSource has been created.

sebastian-peter (changes requested):
Thanks for taking the initiative. Some short questions from my side here. Also, the test for PSW needs to be adapted once this is done, of course.

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
sebastian-peter: Possibly solved [here](https://github.com/ie3-institute/PowerSystemDataModel/pull/999) or with an earlier PR. We should still check if `DateTimeParseException` is also wrapped in a `SourceException` within PSDM

PR #484, opened 2023-03-10
Fix unreliable tests that sometimes fail (with timeouts)

### LINKED ISSUE #359: Fix tests that sometimes fail
Some tests usually succeed, but sometimes don't (e.g. under high cpu load, some timeouts are set too low).

sebastian-peter:
Resolves #359


danielfeismann (approved):
LGTM thanks for your work :)

PR #501, opened 2023-04-06
Fix broken style of references by separating publications and references

### LINKED ISSUE #500: Fix style of references in RTD

danielfeismann:
resolves #500


sebastian-peter (commented):
Thanks for the fix. However, both pages still seem broken at the top right... Maybe there's also a solution to this?

grafik

sebastian-peter:
Somehow, the footer moved up a level in DOM hierarchy. Should normally be somewhere within the `wy-grid-for-nav`-element

grafik


I think that's what's caused the error in the first place

danielfeismann:
Closed, temp solved by #504 and tracked by #505

PR #516, opened 2023-04-27
Align determination of nextState at ChpModel and HpModel

### LINKED ISSUE #515: Have `ChpModel` make use of the new model states
I noticed that the HpModel calls `calculateNextState` in the method `calculateActivePower` while the `ChpModel` doesn't. Since both are supposed to stateful maybe that should happen the same way in both models?

_Originally posted by @t-ober in https://github.com/ie3-institute/simona/pull/82#discussion_r1168828892_

danielfeismann:
resolves #515

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/participant/ChpModel.scala
+++ b/src/main/scala/edu/ie3/simona/model/participant/ChpModel.scala
@@ -78,8 +78,10 @@ final case class ChpModel(
     */
   override protected def calculateActivePower(
       chpData: ChpRelevantData
-  ): ComparableQuantity[Power] =
+  ): ComparableQuantity[Power] = {
+    calculateNextState(chpData)
     chpData.chpState.activePower
+  }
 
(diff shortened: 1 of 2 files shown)


t-ober:
Will not be worked on for know. Address when implementing the CHPAgent or when the updated HPAgent is merged.

src/main/scala/edu/ie3/simona/model/participant/ChpModel.scala
```diff
@@ -78,8 +78,10 @@ final case class ChpModel(
     */
   override protected def calculateActivePower(
       chpData: ChpRelevantData
-  ): ComparableQuantity[Power] =
+  ): ComparableQuantity[Power] = {
+    calculateNextState(chpData)
```
t-ober, line 82: This seems weird since we don't do anything with the state here.

sebastian-peter:
Obsolete due to #1156

PR #519, opened 2023-05-02
Included previousResults in BaseData. Adapted handleResult function in a way that results are only updated when changes occur.

### LINKED ISSUE #487: Only persist switch result when switch position changes
Currently we write out switch results at every pf calculation, and since there are no electrical results for switches this does not make much sense wrt to our discrete event approach. We should only write out switch results whenever the switch position changes.

vickybung1:
Resolves #487


staudtMarius:
This check/filter should be integrated into the new `ResultServiceProxy`.

staudtMarius:
Will be resolved by #1638.

PR #523, opened 2023-05-09
Changed implementation of actor naming 

### LINKED ISSUE #103: Fix unique generation of ActorNames
On very large datasets the current implementation of the actor naming uuid tends to be not unique anymore. Therefore we need to find an alternative for the unique name generation. 

**Possible approaches:**
- remove `substring(0,8)` (might clutter logs...)
- find another method to generate unique identifiers (maybe hash values?

**Affected line:**
Clipboard - 19  Januar 2022 17_03

vickybung1:
Resolves #103


sebastian-peter:
As hinted to in #103, I would like an actor name generation better that is using more meaningful names (not less, as the generation time in nanoseconds has no meaning to the actor). Maybe we can use system participant ids here, if possible?

sebastian-peter:
!test

sebastian-peter (changes requested):
I think the random number wouldn't even be necessary per [this PR](https://github.com/ie3-institute/PowerSystemDataModel/pull/997/files#diff-cb2052bf528a425dff6d9928c937e5914454ef3a0d0fd776dfbcca06a11ed707), but it doesn't hurt either.
Could you take care of the code smell before we merge this @danielfeismann ?

sebastian-peter (approved):
Looks good, let's finally get this in

PR #491, opened 2023-03-17
Introduce squants

danielfeismann:
resolves #490 

This PR is build up mainly on work of @sebastian-peter in #338

PR #555, opened 2023-06-21
Squants in powerflow and grid

### LINKED ISSUE #554: Squants for powerflow and grid
grafik

              Why don't we just use squants in the fold command instead of converting back to quantities?

_Originally posted by @t-ober in https://github.com/ie3-institute/simona/pull/491#discussion_r1188547489_
            
grafik

              We just should have an `.average` implementation for squants. Converting to quantities to average and then convert back to squants kind of defeats the purpose

_Originally posted by @t-ober in https://github.com/ie3-institute/simona/pull/491#discussion_r1188562063_

grafik

              same here

_Originally posted by @t-ober in https://github.com/ie3-institute/simona/pull/491#discussion_r1188563720_

danielfeismann:
resolves #554 
Merge #491 and #561 before


src/test/scala/edu/ie3/simona/model/grid/Transformer3wModelSpec.scala
```diff
@@ -251,16 +237,10 @@ class Transformer3wModelSpec
...
-            Quantities.getQuantity(3.185595567e-6, PU),
-            1e-8
-          )
-          x should equalWithTolerance(
-            Quantities.getQuantity(556.0941828e-6, PU),
-            1e-8
-          )
-          g shouldBe Quantities.getQuantity(0d, PU)
-          b shouldBe Quantities.getQuantity(0d, PU)
+          r =~ Each(3.185595567e-6)
```
danielfeismann, line 240: @sebastian-peter could you have a look here, why this is not identified as false, while `Math.abs(r.toEach - 3.185595567e-6) < 1e-8 shouldBe true` works as expected?

danielfeismann: Edit: With the fix within Transformer3wTestData it is now necessary to change the value (e.g. 1.0e-1) to have same behaviour (in the meaning of result is wrong but test does not identify this)

t-ober (approved):
Nice thanks

PR #558, opened 2023-06-28
Handle codacy issues

### LINKED ISSUE #557: Fix some codacy issues

danielfeismann:
resolves #557

### DIFF
--- a/src/test/groovy/edu/ie3/simona/model/participant/ChpModelTest.groovy
+++ b/src/test/groovy/edu/ie3/simona/model/participant/ChpModelTest.groovy
@@ -6,6 +6,12 @@
 
 package edu.ie3.simona.model.participant
 
+import static edu.ie3.util.quantities.PowerSystemUnits.*
+import static tech.units.indriya.quantity.Quantities.getQuantity
+import static tech.units.indriya.unit.Units.CUBIC_METRE
+import static tech.units.indriya.unit.Units.PERCENT
+import static edu.ie3.util.quantities.QuantityUtil.equals
+
 import edu.ie3.datamodel.models.OperationTime
 import edu.ie3.datamodel.models.StandardUnits
 import edu.ie3.datamodel.models.input.OperatorInput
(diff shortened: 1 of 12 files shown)


src/test/groovy/edu/ie3/simona/model/participant/ChpModelTest.groovy
```diff
@@ -113,16 +112,16 @@ class ChpModelTest extends Specification {
...
-    chpStateNotRunning | 90         | 0           || 0                        // tests case (false, false, true)
-    chpStateNotRunning | 90         | 8 * 115     || 95                        // tests case (false, true, false)
-    chpStateNotRunning | 90         | 10          || 0                        // tests case (false, true, true)
-    chpStateRunning    | 90         | 0           || 95                        // tests case (true, false, true)
-    chpStateRunning    | 90         | 8 * 115     || 95                        // tests case (true, true, false)
-    chpStateRunning    | 90         | 10          || 95                        // tests case (true, true, true)
-    chpStateRunning    | 90         | 7 * 115 + 1 || 95                        // test case (_, true, false) and demand covered together with chp
-    chpStateRunning    | 90         | 9 * 115     || 95                        // test case (_, true, false) and demand not covered together with chp
-    chpStateRunning    | 92         | 1           || 95                        // test case (true, true, true) and storage volume exceeds maximum
```
sebastian-peter, line 125: Are you sure we want to remove the double pipe separators? I remember them as the demarcation between test inputs (on the left side) and expected solutions (on the right side). I think they have no technical meaning, but they just signify the proposed usage of the variables to the developer.

danielfeismann: Technicaly no but I got annoyed by the codacy issue that is raised by this. So another way would be to adapt the codacy pattern, but I'm not sure if there can be some opt-out just for the groovy tests. In my point I would relinquish to have a demarcation line.

sebastian-peter: Hmm well the double pipe is described for the spock testing framework [here](https://spockframework.org/spock/docs/2.3/data_driven_testing.html#_syntactic_variations). 
Codacy thinks the double pipe is a boolean operator here (which it could be in other contexts, but not here), so I think the error is with Codacy. I just disabled this particular check for Groovy files, if you're fine with it I'd just keep it like this, especially because Codacy in my view is just a supplementary code checker besides SonarQube.

danielfeismann: Fine for me, wasn't aware that we can filter these tests based on the type of the file.

src/test/groovy/edu/ie3/simona/model/participant/PvModelIT.groovy
```diff
@@ -55,8 +57,8 @@ class PvModelIT extends Specification implements PvModelITHelper {
     setDefault(US)
 
     pvModels = createPvModels()
-    weatherMap = getWeatherData()
-    resultsMap = getResultsData()
+    weatherMap = weatherData
+    resultsMap = resultsData
```
sebastian-peter, line 61: Honestly, I hate that groovy allows this and Codacy enforces this. But we were already halfway there anyway, so at least let's commit fully

sebastian-peter (approved):
Looks good, thanks

PR #561, opened 2023-07-03
fixing RefSystemTest

### LINKED ISSUE #560: Fix RefSystemTest
`testTolerance < Math.abs(actual.getValue().doubleValue() - expected.getValue().doubleValue())` should be changed to (actual - expected) < testTolerance

danielfeismann:
resolves #560

### DIFF
--- a/src/test/groovy/edu/ie3/simona/model/grid/RefSystemTest.groovy
+++ b/src/test/groovy/edu/ie3/simona/model/grid/RefSystemTest.groovy
@@ -42,7 +42,7 @@ class RefSystemTest extends Specification {
     QuantityUtil.isEquivalentAbs(refSystem.nominalImpedance(), Quantities.getQuantity(166.6666666666666666666666666666666d, OHM), 1E-10)
   }
 
-  def "An dimensionless impedance is transferred correctly between to reference system"() {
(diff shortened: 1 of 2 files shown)


src/test/groovy/edu/ie3/simona/model/grid/RefSystemTest.groovy
```diff
@@ -47,28 +47,28 @@ class RefSystemTest extends Specification {
...
     when:
     Quantity<Dimensionless> actual = RefSystem.transferImpedance(impedance, from, to)
 
     then:
     actual.getUnit() == impedance.getUnit()
-    testTolerance < Math.abs(actual.getValue().doubleValue() - expected.getValue().doubleValue())
+    Math.abs(actual.getValue().doubleValue() - expected.getValue().doubleValue()) < testTolerance
   }
 
   def "An dimensionless admittance is transferred correctly between to reference system"() {
```
sebastian-peter, line 60: "between to" looks like a slip-up... Could be "between reference systems". Would you correct this, also in the test title above?

sebastian-peter (dismissed):
Unreal that we had these broken tests in here without noticing... Thanks for fixing this!

I have only a small wish for another small fix in there.

julianhohmann (approved):

PR #652, opened 2023-11-14
Simplified scheduling protocol

### LINKED ISSUE #650: Introduce new scheduling protocol
The new scheduling protocol includes a bunch changes:
- activations are scheduled without trigger, there is only a simple activation with the current tick
- completions only include the ActorRef of the actor that is completed (no more trigger id) and possibly a new tick

sebastian-peter:
Resolves #650

Most notably, replacing all triggers with `Activation(tick)`. Thus, all actors are ultimately only activated with a tick (which they for the most part only considered anyways).
Actors that are not yet converted to the typed format, are now (where possible) receiving their initialization data within their constructors. This is of course not ideal conceptually, but an ad-hoc solution until they are converted to typed actors later.

Merge first:
- #592
- #654


src/test/scala/edu/ie3/simona/agent/participant/HpAgentModelCalculationSpec.scala
```diff
@@ -678,17 +542,11 @@ class HpAgentModelCalculationSpec
       weatherService.send(hpAgent, RegistrationSuccessfulMessage(Some(0L)))
 
       /* I'm not interested in the content of the CompletionMessage */
-      scheduler.expectMsgType[CompletionMessage]
+      scheduler.expectMsgType[Completion]
       awaitAssert(hpAgent.stateName shouldBe Idle)
 
       /* Send out an activity start trigger */
```
danielfeismann, line 548:
```suggestion
      /* Send out an activation*/
```

src/test/scala/edu/ie3/simona/agent/participant/WecAgentModelCalculationSpec.scala
```diff
@@ -654,17 +510,11 @@ class WecAgentModelCalculationSpec
       weatherService.send(wecAgent, RegistrationSuccessfulMessage(Some(900L)))
 
       /* I'm not interested in the content of the CompletionMessage */
-      scheduler.expectMsgType[CompletionMessage]
+      scheduler.expectMsgType[Completion]
       awaitAssert(wecAgent.stateName shouldBe Idle)
 
       /* Send out an activity start trigger */
```
danielfeismann, line 516:
```suggestion
      /* Send out an activation */
```

src/test/scala/edu/ie3/simona/event/listener/RuntimeEventListenerSpec.scala
```diff
@@ -131,9 +136,9 @@ class RuntimeEventListenerSpec
           s"Simulation until ${calcTime(currentTick)} completed."
         ),
         (
-          Done(endTick, duration, 0, errorInSim = false),
+          Done(endTick, duration, errorInSim = false),
           Level.INFO,
-          s"Simulation completed with \u001b[0;32mSUCCESS (Failed PF: 0)\u001b[0;30m in time step ${calcTime(endTick)}. Total runtime: 3h : 0m : 5s "
+          s"Simulation completed with \u001b[0;32mSUCCESS (Failed PF: 2)\u001b[0;30m in time step ${calcTime(endTick)}. Total runtime: 3h : 0m : 5s "
```
danielfeismann, line 141: Is this really completed with SUCCESS if there are failing PFs. Perhaps we should discuss and resolve this in another thread, as it wasn't caused by your changes..

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
```diff
@@ -166,27 +140,25 @@ abstract class ParticipantAgent[
...
-        triggerId,
         modelBaseStateData
       )
 
     case Event(
-          TriggerWithIdMessage(ActivityStartTrigger(currentTick), triggerId),
+          Activation(currentTick),
           fromOutsideBaseStateData: FromOutsideBaseStateData[M, PD]
         ) =>
       /* An activity start trigger is sent, but I'm still expecting primary data. Go to HandleInformation and wait for
```
danielfeismann, line 158:
```suggestion
      /* An activation is sent, but I'm still expecting primary data. Go to HandleInformation and wait for
```

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
```diff
@@ -283,19 +255,16 @@ abstract class ParticipantAgent[
...
 
     case Event(
-          TriggerWithIdMessage(
-            ActivityStartTrigger(activationTick),
-            triggerId
-          ),
+          Activation(currentTick),
           stateData: DataCollectionStateData[PD]
         ) =>
       /* The actor received an ActivityStartTrigger. Check, if there is everything at its place. If so, change state
```
danielfeismann, line 261:
```suggestion
      /* The actor received an activation. Check, if there is everything at its place. If so, change state
```

danielfeismann (changes requested):
Just very minor things. In addition, there are still ActivityStartTriggers in the puml files.

src/test/scala/edu/ie3/simona/event/listener/RuntimeEventListenerSpec.scala
sebastian-peter: There used to be the setting `simona.time.stopOnFailedPowerFlow`, which now just has a different name. So, we can configure SIMONA to keep running when power flows fail (thus it still can succeed), or we can stop with an error.

sebastian-peter: Ok, just tested this to be sure. 
- If you `simona.powerflow.stopOnFailure` is set to `true`, the `GridAgent` fails if pf does not converge. (It does so in a kind of ugly way, but it still fails. #660 should deal with this).
- If it's set to `false`, we count failed power flows correctly and end with "SUCCESS".

danielfeismann (approved):
Looks good to me, thanks for working in all the adjustments.

Issue #642, opened 2023-11-06
Remove version number from ascii logo


sebastian-peter:
When SIMONA starts, an ASCII logo is printed to console, including a version number `2.0`:

```
   _____ ______  _______  _   _____       ___    ____ 
  / ___//  _/  |/  / __ \/ | / /   |     |__ \  / __ \
  \__ \ / // /|_/ / / / /  |/ / /| |     __/ / / / / /
 ___/ // // /  / / /_/ / /|  / ___ |    / __/_/ /_/ / 
/____/___/_/  /_/\____/_/ |_/_/  |_|   /____(_)____/  
```

 As this could be confused with the release version number, we decided to just remove the number from the logo.

PR #665, opened 2023-11-27
fixing double line in vn_simona config

### LINKED ISSUE #658: vn_simona config double line in comment
```
#########
# ATTENTION: Do not change this config file directly but use it as a base for your personal delta config for the
# vn_simona scenario! Delta configs can be created by including the config you want to change
# parameters from via include <path-to-config> (e.g. include "input/samples/vn_simona/vn_simona.conf") at the
# parameters from via include <path-to-config> (e.g. include "input/samples/vn_simona/vn_simona.conf") at the
# beginning of your config file and then just override the parameters you want to change!
#########
```

line 4

SimonHuette:
resolves #658


sebastian-peter:
@SimonHuette you don't have to, we can just merge #664 first. I can show you another time how to fix this, if you like

sebastian-peter (changes requested):
One more thing that is still missing from this PR (and #642): An addition to `CHANGELOG.md`! If you keep the additions in the same style as what is already there (in the Unreleased section), there's nothing much you can do wrong. Please reach out though if there are any questions.

PR #676, opened 2023-11-29
Introducing scalafix

### LINKED ISSUE #661: Automate imports sorting
It is quite annoying that imports have to be sorted "manually", since different import orders can be committed and make merge conflicts more likely.

Luckily, there are tools that can
1. format and sort imports via gradle command
2. check for correctly formatted imports and fail CI otherwise

Scalafix with the [OrganizeImports rule](https://scalacenter.github.io/scalafix/docs/rules/OrganizeImports.html) is one of those tools. There are further rules like DisableSyntax that could be helpful for keeping the code clean. There is a [gradle plugin for scalafix](https://github.com/cosmicsilence/gradle-scalafix).

Before integrating Scalafix into the project, research other tools that offer similar functionality - there might be some that are better suited for us.

sebastian-peter:
Resolves #661

In this PR also
- scapegoat is removed, as it won't be compatible with Scala 3
- eol settings are set with `.gitattributes`. Recommended setting for `git config core.autocrlf` is `input`
- some more checks than just import order are enabled with scalafix, such as disallowing while loops or redundant syntax
- the changes made via scalafix are also enforced via CI


sebastian-peter:
Closing this one, as we probably won't merge this anytime soon. But for reference, my additions (`.scalafix.conf` etc.) might still be helpful at some future point in time.

PR #682, opened 2023-12-04
Switching to `scala 3`.

### LINKED ISSUE #53: Switch to scala 3 + java 17
Scala 3 has several improvements over the currently used scala version. Even more, in the mid-term I expect that we will run into several problems if we stick to scala 2.

We may consider using this version change to also switch to the latest java LTS which is currently java 17.

For a full switch, we depend on updates of the following tools to support scala 3 
- [x] [Scoverage](https://github.com/scoverage/scalac-scoverage-plugin/issues/299)
- [x] [Scapegoat](https://github.com/scapegoat-scala/scapegoat/issues/521)

We should also consider using this change for a new, actually the first public, release.
Therefore it might make sense to solve #17 before.

staudtMarius:
Resolves #53


sebastian-peter:
As a summary of our discussion: There's a problem with `ParticipantAgent` and `BaseStateData` that involves "applied types" (a scala thing) and possibly implicits. Since the PA is a bit of a mess itself, it'll probably easier if we wait for the rewrite in pekko typed.

PR #720, opened 2024-02-01
Create squants scalatest matchers

### LINKED ISSUE #715: Create squants scalatest matchers
When testing actual result values agsint expected values, we want to disregard rounding errors and thus approximately compare with a tolerance. 

With squants, a common solution currently looks like this: 

```scala
(actualResult~= expectedResult) shouldBe true
```

Comparing values like this has the problem of not giving relevant information on failure. Instead, we see this:

```
Message: false was not equal to true
```

What we rather want here is a matcher equivalent to `QuantityMatchers` for javax quantities.

As always, tests should be provided for the new functionality.

staudtMarius:
Resolves #715

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Instantiation of Heat Pump Agents [#253](https://github.com/ie3-institute/simona/issues/253)
 - Output of accompanying thermal result models
 - Added JDK installation, Scala Plugin + SDK in usersguide [#324](https://github.com/ie3-institute/simona/issues/324)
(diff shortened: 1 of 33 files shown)


sebastian-peter (changes requested):
This looks quite nice already, however (as I said in the issue) it would be nice to have an equivalent to the existing `edu.ie3.simona.test.matchers.QuantityMatchers`. Instead of `equalWithTolerance`, the method should be called `approximate`, which is a bit shorter.

But in the end, comparisons should look like this:
```scala
actualResult should approximate expectedResult
```

This means that `tolerance` needs to be an implicit parameter, the same way that `squants.Quantity#approx` (and therefore `~=`) works.

sebastian-peter (approved):

PR #723, opened 2024-02-05
Solving some compiler warnings.

### LINKED ISSUE #657: Solve some compiler warnings
There's some compiler warnings that are not critical, but could count as "code smells". Not all of them are worth it solving (or can easily be solved), but I believe the ones of the following types can be:
- BigDecimalDoubleConstructor
- "reference to $x is ambiguous"
- "match may not be exhaustive"

Furthermore, some deprecations (gzip, withFirstRecordAsHeader) could possibly be solved (although this might take some more tinkering).

staudtMarius:
Resolves #657


sebastian-peter (approved):
LGTM, thanks!

Issue #646, opened 2023-11-07
Convert groovy tests to ScalaTest, remove groovy


sebastian-peter:
Convert all groovy tests to ScalaTest, then remove everything related to groovy:
- dependencies (spock, ...)
- compile steps, if applicable
- code formatting (spotlessGroovy)
- ...

sebastian-peter:
Tests that are still open to conversion:
- ChpModelTest
- PvModelIT (could be reworked a bit, currently feels a bit "hacky")
- StorageModelTest
- SystemParticipantTest
- CylindricalThermalStorageTest

pierrepetersmeier:
I'm on it :) It's still a bit buggy here and there. Thanks for the update, it's gotten a little confusing.

sebastian-peter:
@pierrepetersmeier No worries, just tried to provide an overview for everyone. If you feel like working on something else, just let us know.

pierrepetersmeier:
All good, thank you for the update. :)

pierrepetersmeier:
Tests that are still open to conversion:

    PvModelIT (could be reworked a bit, currently feels a bit "hacky")
    StorageModelTest
    CylindricalThermalStorageTest

PR #726, opened 2024-02-07
Sh/#646 rewrite groovy test to scala test

SimonHuette:
Here is the finished version of PVModelTest.scala. 

This PR is related to Issue #646


src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
```diff
@@ -0,0 +1,802 @@
...
+      val testCases = Table(
+        ("latitude", "delta", "omegaSSSol"),
+        (0.9d, -0.402449d, 1.0045975406286176d), // lat: ~51.57°N
+        (0.935d, -0.402449d, 0.956011693657339d), // different lat: ~53.57°N
+        (0.9d, 0.017908d, 1.5933675693198284d), // different delta
+        (
+          0.157952297d,
+          0.384670567d,
+          1.635323424114512d,
+        ), // Example 2.2 Goswami Priciples of Solar Engineering
```
sebastian-peter, line 292: This is supposed to be `Principles` I suppose (was a typo before, could you fix nonetheless?) Same at multiple places in this test 😆

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
```diff
@@ -0,0 +1,802 @@
...
+            Degrees(alphaEDeg),
+            Degrees(omegaDeg),
+          )
+
+          // Then
+          thetaG.toDegrees shouldEqual thetaGOut +- 1e-10
+      }
+    }
+
+    "deliver equal angles of incidence for special case" in {
```
sebastian-peter, line 623: Can we find a more expressive title here?

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
```diff
@@ -0,0 +1,802 @@
...
+        (-15d, -7.15, -82.5d, 0d, 0d,
+          80.94904834048776d), // same test but 15° South with 15° less sloped surface
+        (-30d, -7.15, -82.5d, 15d, 180d,
+          80.94904834048776d), // same test but 15° South with 15° more sloped surface (Surface is now facing north, since it is in the southern hemisphere, therefore the surface azimuth is 180°)
+        (52.3d, 23.4337425, 2.15114395d, 0d, 0d,
+          28.91315041538251d), // Berlin 21.06. 12:00 => thetaG = 90 - alphaS
+        (70.3d, 23.4337425, 2.15114395d, 18d, 0d,
+          28.91315041538251d), // same test but 18° North with 18° sloped surface(40d, -11.6d, 82.5d, 60d, 0d, 79.11011928744357d),
+        (40d, -11.6d, 82.5d, 60d, 0d, 79.11011928744357d),
+        (40d, -11.6d, -78.0d, 60d, 0d, 74.92072065185143d),
```
sebastian-peter, line 649: Thanks for merging the tests. I'd still leave in the tests with "inverted omegas" that you left out for clarity though...

sebastian-peter (changes requested):
Some minor remarks from my side that are (hopefully) easy to fix. Good work!

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
```diff
@@ -0,0 +1,800 @@
...
+
+    "provide reactive power up to 110% of its rated apparent power" in {
+      val testCases = Table(
+        ("pVal", "qSol"),
+        (9.5d, 0.004601059995959599d), // above sRated (no q limitation)
+        (11d, 0d), // above sMax (limit q becomes active)
+      )
+
+      forAll(testCases) { (pVal, qSol) =>
+        Given("default adjusted voltage of 1 p.u.")
```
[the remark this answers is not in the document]
SimonHuette: I wasn't sure what to do in the functions where there was the term "expect". I adapted them, also using the GivenWhenThen terms.

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
SimonHuette: What do you think of this title?
"deliver equal angles of incidence at different latitudes adjusted by different slope angles"

danielfeismann: Since the angle of incidence is not always equal over the tests, i would like to avoid the word equal here. But apart from that I like that one.
So maybe `deliver correct angles of incidence at different latitudes and slope angles` ?

Great that this test combines now content which have been in multiple tests before. Maybe we can include some ScalaDoc here giving some context what is covered here. Like the information given in the old test names e.g.: `Testing the equality of zenith angle of a horizontal surface and thetaG of a sloped surface` and maybe the sources as well: e.g. `Iqbal Figure 1.6.2...`

SimonHuette: I agree with your idea for the title, Daniel

SimonHuette: What exactly do you mean by "ScalaDoc"? In the function "deliver correct angles of incidence at different latitudes and slope angles", the information from the old test is preserved in a comment. 
Regarding the sources, I copied them from the old test (wherever possible), however, there are some cases with unknown source (for example in the old "calculate Rb..." function)

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
SimonHuette: I readded the inverted omegas. I also moved them to the function "calculate angle of incidence", since they shouldn't be in the ´deliver correct angles of incidence at different latitudes and slope angles´ function in my opinion

src/test/scala/edu/ie3/simona/model/participant/PvModelSpec.scala
SimonHuette: I added ScalaDoc whenever there were multiple lines of comments. I also added one reference to the Duffie example. Please let me know what you think

SimonHuette:
I noticed that the constructor for PvInput is no longer working here. I looked it up and think it is because of the input parameter "EmInput" which is missing. Is that something that was only recently added to the PvInput class?

danielfeismann:
Yes, this change was introduced with #751. Before the em was told which entities he controlls. We changed this, so now the entities, here the PV, needs to get an em (or none, if the pv is not em controlled). However, it should work if you add `null`, you can compare with `PvModelTest`.

danielfeismann (changes requested):
Good work, I guess we're already done! I only miss a changelog entry and since all tests are transfered you could delete `PvModelTest.groovy`

danielfeismann (approved):
Looks very good to me, thanks for your work done here.

Issue #753, opened 2024-02-27
New Test Grid for Simona 


pierrepetersmeier:
Develop Realistic Grid Operation for Simona

PR #754, opened 2024-02-27
New Test Grid for Simona

pierrepetersmeier:
Resolves : #753

New Grid
CHANGELOG.md


danielfeismann:
Some first points, have still to check this more detailed:
- when publishing a branch the CI/CD pipeline is checking for a certain naming pattern `<initials>/#<issueId>-<description>` looks like you missed a `#` infront of the issue number. Thus the pipeline couldn't be build. Maybe you push it again with corrected branch name.
- maybe you can adapt the coordinates for the line that is still not matching with distance of its coordinates

Note for self: Still have to check results.

pierrepetersmeier:
Okay, I'll check that right away. I want to solve the issue with the cable :).

Issue #644, opened 2023-11-06
Check and correct line length in vnSimona


danielfeismann:
Log statement of vnSimona runs contian several statements concerning line length beeing 50% greater than distance. This should be adapted.

```
13:56:29.454 WARN  e.i.d.u.validation.ValidationUtils - Line length is more than 50.0% greater than the calculated distances between points building the line: LineInput{uuid=63be5ffd-acbc-4c4f-aab9-54c616e2dd2e, id=LTG_NS_NET126_F4_(9)-NS_NET126_F4_(10), operator=a2c896f4-d661-4248-89ff-99e94d08773e, operationTime=OperationTime{startDate=null, endDate=null, isLimited=false}, nodeA=119d270a-ff22-4fdb-8214-cb5b336790bf, nodeB=857c264a-7072-4bb7-af56-2f01539b2a2e, noOfParallelDevices=1, type=a4aab125-8af4-4ef9-8842-4cf718f998bc, length=0.032999999821186066 km, geoPosition=LINESTRING (7.4116482000001005 51.4843281000001, 7.4116482 51.4843281), olmCharacteristic=OlmCharacteristicInput{points=[CharacteristicCoordinate{x=0.0 m/s, y=1.0 p.u.}]}}

```

danielfeismann:
Other warnings might be checked as well:

```
Entity may be unsafe because of: Measurement Unit does not measure any values [MeasurementUnitInput{uuid=b400da1c-8f46-463a-92aa-28ec48391a7b, id=regulate_transformers, operator=a2c896f4-d661-4248-89ff-99e94d08773e, operationTime=OperationTime{startDate=null, endDate=null, isLimited=false}, node=15345698-c319-461f-b969-37d50fb84220, vMag=false, vAng=false, p=false, q=false}]
 
```

sebastian-peter:
> ```
> Entity may be unsafe because of: Measurement Unit does not measure any values
> ```

These ones are fixed in #672

pierrepetersmeier:
@danielfeismann The Warning is not coming from vnSimona but from PSDM, specifically in ConnectorValidtionUtils.Java at line 608.

pierrepetersmeier:
I haven't fully understood the code, but my assumption is as follows: In the default line_input.csv file, all nodes receive the default geoposition but have different lengths. When the lengths of the lines are checked against the nodes, they don't match up.

pierrepetersmeier:
[...]

PR #757, opened 2024-02-28
New test grid for simona

### LINKED ISSUE #753: New Test Grid for Simona

pierrepetersmeier:
Resolves: #753

New Branch publishing, because false naming pattern.
Thanks for the tip @danielfeismann


pierrepetersmeier:
The line where the warning occurs has the same coordinates as in #644, i.e. the "distance" 0.

danielfeismann:
> Okay, if we change the coordinates minimally and they are no longer exactly the same, the warning disappears.

Would be fine for me

pierrepetersmeier:
should I do the same in vn_Simona? Maybe it's not quite as clean because there's more than one line, but then the warnings are gone.

pierrepetersmeier:
> > Okay, if we change the coordinates minimally and they are no longer exactly the same, the warning disappears.
> 
> Would be fine for me

should I do the same in vn_Simona? Maybe it's not quite as clean because there's more than one line, but then the warnings are gone.

danielfeismann:
Since I guess it wouldn't fit to often it will not help that much at vn_simona. But if you want give it a try, feel free.

pierrepetersmeier:
> Since I guess it wouldn't fit too often it will not help that much at vn_simona. But if you want give it a try, feel free.

Yes, that didn't work :/

pierrepetersmeier:
> This looks very good already @pierrepetersmeier. Thanks for the steps done here, it made it very easy for me to make the following adjustments: 
> - moved the grid to the TU Dortmund side
> - moved both WECs to the upper voltage level to reduce the voltage rise in the lv grid 😄 
> - changes in config (changes from main branch and write output files)
> 
> More things could be done (e.g. include heat pumps). But I think we should stick with this for now.
> Will bring this into our internal discussions if it suits our needs and will get back to you with the result.

alright thank you :)

PR #765, opened 2024-03-07
New scaling of participants

### LINKED ISSUE #734: Consider scaling factor with flex options
Scaling is applied in `SystemParticipant.calculateApparentPower`, which is not utilized by participant models when calculating flex options and handling flex control.

### LINKED ISSUE #764: Implementing scaling of all relevant input parameters
Instead of just scaling the output apparent power of system participants, we should rather scale all relevant input parameters (including capacity etc.). This is made possible by https://github.com/ie3-institute/PowerSystemDataModel/pull/1012 and the release of PSDM 5.0.

Linked to #734, which is also solved by this.

sebastian-peter:
Resolves #734 
Resolves #764


danielfeismann (approved):
LGTM

PR #770, opened 2024-03-11
realistic stochastic load agent

### LINKED ISSUE #691: Conceptualize and Implement a "realistic" load agent
We are currently lacking a load agent that is "realistically" stochastic. Our current Agents consist of

1. `ProfileLoadAgent`: Which represents _standard load profiles_. Which is only really realistic when looking at a multitude of loads in that respective category, since it basically depicts their average behavior. 
2. `FixedLoadAgent`: Which has a fixed load.
3. `RandomLoadAgent`: Which is kind of what we want, but not really. The problem here is, that the household picks a random value from a distribution. That means the previous value has nothing to do with the last value (we probably should double check, that that is in fact true). 

We need a household that is stochastic, but without the limitation of the `RandomLoadAgent`, meaning that not only the individual values depict a expected distribution, but also the sequence of values is sensible. One modeling approach could be to go for a Markov chain type of process, similar as our half-finished implementation [here](https://github.com/ie3-institute/MarkovsHousehold).

pierrepetersmeier:
Resolves: #691


staudtMarius:
This branch can be closed, because it is not necessary to do the implementation in SIMONA.

Issue #375, opened 2022-10-12
Remove documentation index page


sebastian-peter:
The links under "Indices and tables" on the [index page](https://simona.readthedocs.io/en/latest/) are broken.

While fixing this, one could also try to completely migrate from rst to markdown (only the index and [reference page](https://simona.readthedocs.io/en/latest/references.html) are left). Then, maybe (and only maybe) one could also remove rst support, slimming down the documentation setup a bit.

Useful links:
- [Directives in myst](https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html#roles-directives)
- [References with myst](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#markdown-syntax-using-myst)

sebastian-peter:
For now, just fixing the dead links is enough for this issue

PR #771, opened 2024-03-11
Removing documentation index page

pierrepetersmeier:
Resolves: #375


docs/readthedocs/allpages.md
```diff
@@ -0,0 +1,17 @@
...
+---
+maxdepth: 1
+---
+.. toctree::
+   :maxdepth: 2
+   :glob:
+
+   *
+   !allpages
+```
```
danielfeismann, line 17: This stacked toctree seems to cause the error
```
/home/docs/checkouts/readthedocs.org/user_builds/simona/checkouts/771/docs/readthedocs/allpages.md:7: WARNING: toctree contains reference to nonexisting document '.. toctree::'
/home/docs/checkouts/readthedocs.org/user_builds/simona/checkouts/771/docs/readthedocs/allpages.md:7: WARNING: toctree contains reference to nonexisting document '   :maxdepth: 2'
/home/docs/checkouts/readthedocs.org/user_builds/simona/checkouts/771/docs/readthedocs/allpages.md:7: WARNING: toctree contains reference to nonexisting document '   :glob:'
/home/docs/checkouts/readthedocs.org/user_builds/simona/checkouts/771/docs/readthedocs/allpages.md:7: WARNING: toctree contains reference to nonexisting document '   *'
/home/docs/checkouts/readthedocs.org/user_builds/simona/checkouts/771/docs/readthedocs/allpages.md:7: WARNING: toctree contains reference to nonexisting document '   !allpages'
```

danielfeismann (changes requested):

docs/readthedocs/allpages.md
pierrepetersmeier: Yes, unfortunately the solution doesn't work because you can apparently only have one toctree. @sebastian-peter and I imagined that we would have an index page, a search page and a page where all pages are listed chronologically (Modindex/allpages). Unfortunately I haven't found any other solution apart from toctree. But I'll keep looking now.

sebastian-peter: @pierrepetersmeier I think we can skip the listing of all pages, if it's not easily implementable. Our main navigation should fit the purpose anyways.

sebastian-peter: Just fyi, I found out what the index is actually supposed to be. You can add index entries on all pages with `.. index:: Entry name`, and they're then added to the index linking back to the page that they appear on. So it's a literal index similar to how it works in books. Since we currently don't use this, it's definitely fine to remove it.

sebastian-peter (approved):

docs/readthedocs/allpages.md
pierrepetersmeier: Alright, good to know, maybe we'll need it sometime.

PR #772, opened 2024-03-13
Print log directory to terminal upon simulation failure

### LINKED ISSUE #626: Print directory of logs to terminal upon simulation failure
It would be nice if we would print the directory location to the terminal if something goes wrong, so the user can quickly navigate to the log file to investigate the issue.

staudtMarius:
Resolves #626


src/main/scala/edu/ie3/simona/main/RunSimona.scala
```diff
@@ -68,6 +70,21 @@ trait RunSimona[T <: SimonaSetup] extends LazyLogging {
...
+      val root: Path = Path.of(
+        new File(".").getAbsoluteFile.getParent,
+        "logs",
+        "simona",
+        "simona.log",
+      )
+      logger.error(
+        s"Simulation stopped due to the occurrence of an error! The full log can be found here: $root"
+      )
+    }
```
t-ober, line 87: This should take into account the actual output directory that was specified in the configuration. It seems like the path you are printing is relative to simona not based on the configuration

t-ober (changes requested):

src/main/scala/edu/ie3/simona/sim/setup/SimonaStandaloneSetup.scala
```diff
@@ -66,6 +66,7 @@ class SimonaStandaloneSetup(
     resultFileHierarchy: ResultFileHierarchy,
     runtimeEventQueue: Option[LinkedBlockingQueue[RuntimeEvent]] = None,
     override val args: Array[String],
+    override val logOutputDir: String,
```
sebastian-peter, line 69: You could also add `override val logOutputDir: String = resultFileHierarchy.logOutputDir` to the class body instead. That way, we'll have one less variable within the class parameters.

sebastian-peter (changes requested):
Small suggestion from my side

t-ober (approved):

sebastian-peter (approved):

PR #779, opened 2024-04-03
Refactoring of `GridAgent` messages.

### LINKED ISSUE #736: Refactoring GridAgent messages
As in #666 commented the different ``GridAgentMessage``s should be collected into one ``object``. Also the messages should be split into requests and responses.

staudtMarius:
Resolve #736


src/main/scala/edu/ie3/simona/agent/grid/GridAgentMessages.scala
```diff
@@ -0,0 +1,189 @@
...
+    *   received response
+    */
+  final case class WrappedResponse(response: GridAgent.Response)
+      extends GridAgent.InternalRequest
+
+  /** Wrapper for [[PowerMessage]]s.
+    * @param msg
+    *   the received power message
+    */
+  final case class WrappedPowerMessage(msg: PowerMessage)
```
sebastian-peter, line 93: There should be no need to wrap power messages. They should be either always received by GridAgents or always received by ParticipantAgents.

For example, `RequestAssetPowerMessage` is always received by a ParticipantAgent. It should thus be part of the messages received by ParticipantAgents. 

`AssetPowerChangedMessage` is always received by GridAgents and should thus extend `GridAgent.Request` (at least transitively).

sebastian-peter (changes requested):

src/main/scala/edu/ie3/simona/agent/grid/GridAgentMessages.scala
```diff
@@ -0,0 +1,266 @@
...
+    */
+  final case class WrappedActivation(activation: Activation)
+      extends GridAgent.InternalRequest
+
+  /** Wrapper for any [[GridAgent.Response]] from [[GridAgent]]s.
+    * @param response
+    *   received response
+    */
+  final case class WrappedResponse(response: GridAgent.Response)
+      extends GridAgent.InternalRequest
```
sebastian-peter, line 91: Hmm. I get why this exists, but there should be a way that is a bit more concise and easier to grasp. If we throw out `WrappedResponse`, and also the trait `Response`, we could then have `InternalResponse` be renamed to `InboundResponse` (or something like that) and extend `Request` as well. This makes sense because `Request` should be extended by everything that we can receive.

sebastian-peter (changes requested):

sebastian-peter (approved):
Looks good, thanks

PR #780, opened 2024-04-04
Making passing of `RefSystem` via config optional.

### LINKED ISSUE #769: Make passing a `refSystem` via the configuration optional
Currently we need to pass a `refSystem` explicitly via the configuration. I don't see why we can't make that optional and use sensible standard values so the user do not have to care about that if they don't want to.

```
  gridConfig {
    refSystems = [
      {
        sNom = "60 MVA"
        vNom = "10 kV"
        voltLvls = [
          {
            id = "MS"
            vNom = "10 kV"
          }
        ]
      }
```

staudtMarius:
Resolves #769


staudtMarius:
!test

staudtMarius:
!test

docs/readthedocs/config.md
```diff
@@ -204,20 +204,28 @@ Tba:
 
 ## Grid configuration 
 
+The reference system contains a list of voltage levels. Each element includes the nominal apparent power, the nominal 
+voltage and the separate configuration of each voltage level. The voltage level configuration is composed of the identifier 
+and the nominal voltage.
+
+The configuration of a reference system is optional. If no configuration is provided by the user, the default
+[reference system](models/reference_system) that includes all common german voltage levels is used. For those users
```
danielfeismann, line 212: Not super important but I think we already use (sphinx-hoverxref)[https://sphinx-hoverxref.readthedocs.io/en/latest/] within the documentation. So we could provide a hovered preview here as well? (This might be moved into a new Issue for some future HiWi)

danielfeismann (approved):
LGTM, thanks for improving SIMONA a little bit more :)

docs/readthedocs/config.md
danielfeismann: Moved into future issue

danielfeismann (approved):
LGTM, thanks for improving SIMONA a little bit more :)

PR #788, opened 2024-04-17
Fix scheduling at evcs with more than one ev at a time without em

### LINKED ISSUE #787: Fix scheduling at evcs with more than one ev at a time without em
In cases where more than one ev charges at an evcs the results for soc and the resulting power of the evcs in the output are wrong. 

grafik

danielfeismann:
resolves #787


src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
```diff
@@ -454,6 +455,9 @@ protected trait EvcsAgentFundamentals
         newState,
       ),
       resultValueStore = updatedResultValueStore,
+      additionalActivationTicks = baseStateData.additionalActivationTicks.union(
```
danielfeismann, line 458: This seems not to work. Any ideas? @sebastian-peter

danielfeismann: Found a way to add the additionalTick.

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
```diff
@@ -466,6 +468,51 @@ protected trait EvcsAgentFundamentals
...
+    // in order to create 0kW entries for EVs that do not
+    // start charging right away at lastTick, create mock
+    // schedule entries that end before lastTick
+
+    val defaultEntry = ScheduleEntry(lastTick, lastTick, Kilowatts(0d))
+
+    val schedule = schedules.getOrElse(ev.uuid, SortedSet(defaultEntry))
+
+    val updatedSchedule = schedule.map { entry =>
+      if (currentTick >= entry.tickStart && currentTick <= entry.tickStop)
```
sebastian-peter, line 507: Why do we filter out schedule entries that start after the current tick? They'll still be valid at some point in the future, right?

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
```diff
@@ -442,19 +443,20 @@ protected trait EvcsAgentFundamentals
...
-        .view
-        .mapValues {
-          // Remove schedules that ended before or at current tick.
-          // Schedule entries ending at current tick do not have any
-          // impact on the schedule from the current tick on
-          _.filter(_.tickStop > tick)
-        }
-        .toMap
+    // filter(_.tickStop > currentTick)
+    // TODO is it possible to remove also the schedules that ended at currentTick? -> probably yes, test required
```
sebastian-peter, line 447: Should be possible because we're done with these events at this point, right?

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
```diff
@@ -2004,6 +2009,396 @@ class EvcsAgentModelCalculationSpec
...
+      }
+
+      // Check the results
+      // expect 7 Evcs Results and 12 Ev Results
+      // FixMe: 7 Evcs Results is correct but the one for tick 5400 is missing, the one at tick 0 comes twice...
+
+      Range(0, 19)
+        .map { _ =>
+          resultListener.expectMsgType[ParticipantResultEvent]
+        }
```
sebastian-peter, line 2255: I think it'd make sense to test for these events for the given times separately, because then we don't need `match` for the right times. So, after results for tick 900 should have been sent out, we're waiting for 4 results, etc.

src/test/scala/edu/ie3/simona/test/common/EvTestData.scala
```diff
@@ -14,19 +14,27 @@ import java.util.UUID
...
-    UUID.fromString("6d7d27a1-5cbb-4b73-aecb-dfcc5a6fb22e"),
+    UUID.fromString("0-0-0-0-b"),
     "evB",
     Quantities.getQuantity(11d, PowerSystemUnits.KILOWATT),
     Quantities.getQuantity(11d, PowerSystemUnits.KILOWATT),
     Quantities.getQuantity(80d, PowerSystemUnits.KILOWATTHOUR),
     200,
   )
+  protected val evC: MockEvModel = new MockEvModel(
+    UUID.fromString("0-0-0-0-c"),
```
sebastian-peter, line 33: I assigned some UUIDs that are recognizable here, was useful when I was debugging fyi

sebastian-peter (changes requested):
Some questions regarding the adapted functionality...

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
```diff
@@ -528,6 +576,12 @@ protected trait EvcsAgentFundamentals
         modelBaseStateData
     }
 
+    determineResultsAnnounceUpdateValueStore(
+      lastState,
+      currentTick,
+      modelBaseStateData,
+    )
```
sebastian-peter, line 583: Also, because I keep having to search for this: *This* is the most important part of the PR, I think, because it fixes results not being generated if several arrivals happen in succession without any departure. Before this PR, results were only generated when EVs depart.

src/test/scala/edu/ie3/simona/test/common/input/EvcsInputTestData.scala
```diff
@@ -43,4 +46,19 @@ trait EvcsInputTestData extends DefaultTestData with NodeInputTestData {
...
+    QControl.apply(new CosPhiFixed("cosPhiFixed:{(0.0,1.0)}")),
+    Kilowatts(100),
+    ElectricCurrentType.AC,
+    1.0d,
+    4,
+    EvcsLocationType.HOME,
+    vehicle2grid = false,
+    ChargingStrategy.MAX_POWER,
+    lowestEvSoc = 0.2,
+  )
```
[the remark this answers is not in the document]
danielfeismann: Good catch. This is no longer used.

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
danielfeismann: Yes, I agree. But wouldn't they reappear in the next tick? I don't remember the details, maybe this change overlapped with some other change, so I got the impression that this was necessary. However, tests show that this change is not necessary at all, so I will rollback here.

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
danielfeismann: This is no longer valid with the rollback above, i guess.

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
```diff
@@ -466,6 +468,51 @@ protected trait EvcsAgentFundamentals
...
+      schedules: EvcsModel.ScheduleMap,
+  ): EvcsModel.ScheduleMap = {
+
+    // in order to create 0kW entries for EVs that do not
+    // start charging right away at lastTick, create mock
+    // schedule entries that end before lastTick
+
+    val defaultEntry = ScheduleEntry(lastTick, lastTick, Kilowatts(0d))
+
+    val schedule = schedules.getOrElse(ev.uuid, SortedSet(defaultEntry))
```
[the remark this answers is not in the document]
danielfeismann: This is no longer valid with the rollback above, i guess.

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
danielfeismann: I find this one easier to understand because if we organise it the other way one always has to jump between the steps since the results for tick 900 will be available at tick 2700. But I'm not insisting on this.

danielfeismann: As discussed I tried to rearrange the results. But I'm not really happier with this solution. It produced ~100 lines more and is in my opinion still less understandable. (But in the end I would be able to accept also this one :) )

src/main/scala/edu/ie3/simona/agent/participant/evcs/EvcsAgentFundamentals.scala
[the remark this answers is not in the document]
danielfeismann: Changed. This is also covered by the test in EvcsAgentFundamentalsSpec

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
sebastian-peter: I prepared a suggestion at two places where results are matched. Would've just done it a lot shorter like this. If the result type does not match, a `MatchError` is thrown, which is fine imo.

danielfeismann: Like the Evcs part very much and also took some of your improvements for Evs for the later test cases. Please have a look

sebastian-peter: Thanks, looks great. With the new ParticipantAgent I intend to send results of the active tick always, so that'd be a bit more easy to grasp in the test.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsModel.scala
```diff
@@ -207,7 +207,7 @@ final case class EvcsModel(
     * @return
     *   The charged EV
     */
-  private def chargeEv(
+  def chargeEv(
```
sebastian-peter, line 210: Can still be private I think

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsModel.scala
```diff
@@ -325,10 +325,10 @@ final case class EvcsModel(
         val time = tick.toDateTime(simulationStartDate)
 
         // separate into those entries that are still active
-        // and those that have ended before or at tick
+        // and those that have started before tick and will end after tick
         val (stillActive, endedEntries) = lastActiveEntries.partition {
           case (_, entry) =>
-            entry.tickStop > tick
+            (entry.tickStart < tick && entry.tickStop > tick)
```
sebastian-peter, line 331: This also seems unnecessary, because entries that start after current tick cannot be present in `lastActiveEntries` anyway.

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
```diff
@@ -1987,6 +1992,425 @@ class EvcsAgentModelCalculationSpec
...
+        case ParticipantResultEvent(result: EvcsResult) =>
+          result.getInputModel shouldBe evcsInputModel.getUuid
+          result.getTime shouldBe 4500.toDateTime
+          result.getP should beEquivalentTo(22d.asKiloWatt)
+          result.getQ should beEquivalentTo(0d.asMegaVar)
+      }
+      /* FixMe: We would expect another Evcs Result for the lastTick of 5400 here.
+         But this can't be calculated since there is no nextTick.
+         For simulation it is as well necessary to fix this e.g. by writing the lastResults when finishing simulation.
+       */
```
sebastian-peter, line 2413: This should be fixed with the new ParticipantAgent, but please just leave this in for now.

sebastian-peter (changes requested):
Hitting the home stretch...

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsModel.scala
danielfeismann: Good catch. We missed that one when reverting that change above

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
```diff
@@ -1987,6 +1992,425 @@ class EvcsAgentModelCalculationSpec
...
+              result.getQ should beEquivalentTo(0d.asMegaVar)
+              result.getSoc should beEquivalentTo(0.asPercent, 1e-2)
+            case model if model == ev2700.uuid =>
+              throw new RuntimeException(
+                s"Ev $model should arrive later at Evcs but it's already there."
+              )
+            case _ =>
+              throw new RuntimeException(
+                s"Unexpected case at EvcsAgentModelCalculationSpec occurred."
+              )
```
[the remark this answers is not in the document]
danielfeismann: Sure? Because when removing those and increase the number of expected results by one just for testing, the test will still be fine but in my opinion it should fail...

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsModel.scala
```diff
@@ -514,13 +514,13 @@ final case class EvcsModel(
       modelState: EvcsState,
       data: EvcsRelevantData,
   ): ApparentPower =
-    throw new NotImplementedError("Use calculatePowerAndEvSoc() instead.")
+    throw new NotImplementedError("Use chargeEv() instead.")
```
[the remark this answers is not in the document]
danielfeismann: Not sure. Wouldn't we like to take the schedule that becomes active when charging the ev? Most certain it will be the same as of `calculateNewScheduling()` but couldn't there be some cases where this not happens?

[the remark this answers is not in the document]
sebastian-peter: In the end, it doesn't matter that much because there's no direct equivalent to the methods above in EvcsModel.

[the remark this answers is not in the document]
danielfeismann: That's true, We could give both as possiblity to use instead?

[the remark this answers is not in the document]
sebastian-peter: Yes why not

src/test/scala/edu/ie3/simona/agent/participant/EvcsAgentModelCalculationSpec.scala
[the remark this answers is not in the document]
sebastian-peter: Could you provide me with how you did that? I tried somehow what you said and it fails...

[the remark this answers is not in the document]
danielfeismann: Mh, I stashed what was failing yesterday, but now it works somehow. My bad.

sebastian-peter (approved):
LGTM, thanks for you patience!

PR #798, opened 2024-04-25
Fix ConfigFailFastSpec after Refsystem passing

### LINKED ISSUE #797: VnSimona breaks with optional RefSystem passing
with changing the voltage level abbreviations to english terms (LV,MV,HV,EHV) when introducing #780 VnSimona breaks.
ConfigFailFastSpec need to be updated as well

danielfeismann:
resolves #797

### DIFF
--- a/src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
+++ b/src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
@@ -114,7 +114,8 @@ case object ConfigFailFast extends LazyLogging {
 
     // check if the provided combinations of refSystems provided are valid
     val refSystems = simonaConfig.simona.gridConfig.refSystems
-    refSystems.foreach(checkRefSystem)
+    if (refSystems.isDefined)
+      refSystems.foreach(refsys => checkRefSystem(refsys))
 
     /* Check all participant model configurations */
     checkParticipantRuntimeConfiguration(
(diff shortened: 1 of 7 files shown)


staudtMarius:
@danielfeismann Is there a reason for changing the config template?

danielfeismann:
Imho, if we want to have the refsystem only optional in config, we need to adapt this. I didn't saw that in my review. But maybe misunderstood something here...

staudtMarius:
The point of #780 was to make the line in the config optional. If the user does not want to specify the refsystem, then a default one should be used. Therefore I adapted the config template.

danielfeismann:
Right. But there are any changes of 'config-template.conf' in #780. Thus I started this PR. Need to fix some more things and will come back to you then.

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
```diff
@@ -431,67 +432,71 @@ case object ConfigFailFast extends LazyLogging {
...
 
-    val voltLvls =
-      refSystem.voltLvls.getOrElse(List.empty[SimonaConfig.VoltLvlConfig])
-    val gridIds = refSystem.gridIds.getOrElse(List.empty[String])
+  private def checkRefSystem(refSystems: List[RefSystemConfig]): Unit = {
+    refSystems.foreach { refSystem =>
+      {
+        val voltLvls =
+          refSystem.voltLvls.getOrElse(List.empty[SimonaConfig.VoltLvlConfig])
+        val gridIds = refSystem.gridIds.getOrElse(List.empty[String])
```
t-ober, line 441: Why do I need to pass a grid id ?
Can't we just define a default refsystem for all sensible voltage level and use that.
My goal is that the user does not even need to worry what a ref system is.

t-ober (changes requested):

src/main/scala/edu/ie3/simona/config/ConfigFailFast.scala
danielfeismann: This will check the grid id when the users provide a RefSystem, just as it does in the past. You want to remove the ability to specify a RefSystem via config completely and not just make it optional?

t-ober: whoops. you are right slightly misread it. that gets only run if a refsystem is actually defined

t-ober (changes requested):
Looks good to me. You might take a short look at the sonarqube complaints. We could at least get rid of that unused variable.

t-ober (approved):

PR #804, opened 2024-04-30
Provide uuids of the unconnected elements

### LINKED ISSUE #799: Provide uuids of the unconnected elements
```
edu.ie3.simona.exceptions.GridInconsistencyException: The grid with subnetNo 1 is not connected! Please ensure that all elements are connected correctly and inOperation is set to true!
```

It would be great for debuging to get the uuids of the unconnected elements.

danielfeismann:
Merge #807 first!

resolves #799


staudtMarius:
With the [PSDM 1092](https://github.com/ie3-institute/PowerSystemDataModel/pull/1092) the `ValidationUtils` will check the connectivity for all defined operation time intervals. It will also put out the unconnected elements.

Therefore this part could be completly removed from SIMONA, because we would than check the connectivity directly after reading the grid.

danielfeismann:
This is already included in the last PSDM update. This PR is therefore obsolete.

PR #807, opened 2024-05-06
Fix GridSpec

### LINKED ISSUE #806: Fix GridSpec
- Pictogram of `FiveLinesWithNodes` doesn't match the lines forming the grid
- GridSpec test `throw an InvalidGridException if a grid is not connected` throws exception but not for the reason planned.

danielfeismann:
resolves #806


jo-bao (approved):
Seems to throw the right exception!

PR #812, opened 2024-05-23
Fix bug in default `RefSystem`.

### LINKED ISSUE #811: Default reference system uses unit `volt` for lv.
image

This leeds to an error here:
image

The error occures when calculating the difference, because one value is given in the unit `Volt` while the other is given in `Kilovolt`.

staudtMarius:
Resolves #811


danielfeismann (approved):
Good catch!

PR #816, opened 2024-05-29
Fix color in log entry after power flow calculation

### LINKED ISSUE #814: Color in Log "Simulation completed" is black
The color after Power Flow Calculation in the log entry is not reseted.

jo-bao:
Resolves #814


jo-bao:
!test

danielfeismann (approved):
Good one!

PR #823, opened 2024-05-31
Fix simulation stopping at unhandled messages in `DBFSAlgorithm`.

### LINKED ISSUE #821: Fix simulation stopping at unhandled messages in `DBFSAlgorithm`
There seems to be an issue in the way the `DBFSAlgorithm` handles `unhandled` messages.

image

staudtMarius:
Resolves #821

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
+++ b/src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
@@ -491,7 +491,7 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
           idle(cleanedGridAgentBaseData)
 
         case (message, _) =>
-          ctx.log.debug(s"Received the message $message to early. Stash away!")
+          ctx.log.debug(s"Received the message $message too early. Stash away!")
           buffer.stash(message)
           Behaviors.same
       }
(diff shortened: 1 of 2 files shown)


src/main/scala/edu/ie3/simona/agent/grid/DBFSAlgorithm.scala
```diff
@@ -802,9 +803,10 @@ trait DBFSAlgorithm extends PowerFlowSupport with GridResultsSupport {
           buffer.stash(powerResponse)
           Behaviors.same
 
-        case _ =>
-          // preventing "match may not be exhaustive"
-          Behaviors.unhandled
+        case (message, _) =>
+          ctx.log.debug(s"Received the message $message to early. Stash away!")
```
sebastian-peter, line 807: same as above

sebastian-peter (changes requested):
Also question from my side: Was this an issue that also existed before the transition to pekko typed?

staudtMarius:
@sebastian-peter I never noticed this issue before the conversion to typed.

sebastian-peter:
@staudtMarius Then let's analyze what the problem is and where it comes from first before merging this

sebastian-peter:
@staudtMarius We still have a todo here if I remember correctly, which was specifying the messages that are supposed to be stashed

sebastian-peter (approved):
Thanks, looks good

PR #825, opened 2024-06-04
Additional tests to check flexibility options of thermal house and storage

### LINKED ISSUE #729: Additional tests to check flexibility options of thermal house and storage
- e.g. thermal house without storage
- ...

based on this comment:

     Should we also have tests for the other cases?

_Originally posted by @danielfeismann in https://github.com/ie3-institute/simona/pull/714#discussion_r1482582742_

danielfeismann:
resolves #729


src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -231,55 +231,186 @@ class HpModelSpec
...
+            (
+              ThermalGridState(
+                Some(ThermalHouseState(0L, Celsius(15), Kilowatts(0))),
+                Some(ThermalStorageState(0L, KilowattHours(20), Kilowatts(0))),
+              ),
+              HpState(
+                isRunning = false,
+                0,
+                Some(hpData.ambientTemperature),
+                Kilowatts(0.0),
```
jo-bao, line 250: Why is the expression changed from the double value?

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -231,55 +231,186 @@ class HpModelSpec
...
+                Kilowatts(80.0),
+                ThermalGridState(
+                  Some(ThermalHouseState(0L, Celsius(22), Kilowatts(80))),
+                  Some(
+                    ThermalStorageState(0L, KilowattHours(500), Kilowatts(0))
+                  ),
+                ),
+                Some(HouseTemperatureUpperBoundaryReached(0L)),
+              ),
+              (0.0, 0.0, 0.0),
```
jo-bao, line 371: Short question: The case that the house temperature is above 22°C is not possible because we don't consider heating from the environment?

jo-bao (commented):
Over all it looks good to me! Just some minor points.

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: So far it should be same case as ==22°C but I'll add this case as well.

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: If you refer to the case lefthand where the lastHouseState was 21°C and 80 Kilowatts feedin, this one should still be in the table (labeled `Storage and house have remaining capacity`)

docs/readthedocs/models/thermal_grid_model.md
```diff
@@ -3,6 +3,8 @@
 
 The Thermal Grid Model introduces a coupling point to thermal system, equivalent to an electrical node. It can be used to interconnect thermal units to a thermal heat network.
 
+A Thermal Grid Model consists of a {ref}`thermal_house_model` and / or a {ref}`cts_model` which are supplied by a thermal source like a [Heat pump].
```
sebastian-peter, line 6: Link to heat pump does not work. 

Also, I think the thermal grid can really be configured to have no house, but in simulation this does not work, right? Isn't the house then (currently) mandatory?

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -232,55 +232,209 @@ class HpModelSpec
...
-            Some(
-              ThermalHouseState(
-                0L,
-                Celsius(21),
-                Kilowatts(80),
-              )
+    "determining the flexibility options for different states" should {
+      "deliver correct flexibility options" in {
+        val testCases
+            : TableFor3[ThermalGridState, HpState, (Double, Double, Double)] =
```
sebastian-peter, line 238: Nothing important, but question from my side: Does the specification of the type `TableFor3` provide any benefits?

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -232,55 +232,209 @@ class HpModelSpec
...
+                  Some(
+                    ThermalStorageState(0L, KilowattHours(500), Kilowatts(0))
+                  ),
+                ),
+                Some(HouseTemperatureUpperBoundaryReached(0L)),
+              ),
+              (0.0, 0.0, 0.0),
+            ),
+
+            // No capacity for flexibility at all when house has been (external) heated up above upperTemperatureBoundary
```
sebastian-peter, line 375:
```suggestion
            // No capacity for flexibility at all when storage is full and house has been (externally) heated up above upperTemperatureBoundary
```

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -232,55 +232,209 @@ class HpModelSpec
...
+            // Initialize the house and grid models
+            val house =
+              thermalHouse(18, 22).copy(ethLosses = WattsPerKelvin(200))
+            val grid = thermalGrid(house, Some(thermalStorage))
+            val hp = hpModel(grid)
+
+            // Create relevant data for the current test
+            val relevantData = hpData.copy(currentTick =
+              thermalState.houseState.map(_.tick).getOrElse(0L)
+            )
```
sebastian-peter, line 418: Last tick is 0 and current tick is 0. This usually would not happen. I saw that you set `activePower` and `qDot` to different values for this test, but these values are ignored anyways, right? Because, to the current state (once calculated), it's not relevant what the last power has been.

So maybe it'd be clearer to use, let's say, 0 as last tick and 900 as current tick, and just set the powers to 0? Looking through the model, on my side there's still some confusion left (relating to our discussion about `calcState`), so maybe we have to get some clarity there beforehand.

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -232,55 +232,209 @@ class HpModelSpec
...
+                ThermalGridState(
+                  Some(ThermalHouseState(0L, Celsius(19), Kilowatts(0))),
+                  Some(
+                    ThermalStorageState(0L, KilowattHours(20), Kilowatts(0))
+                  ),
+                ),
+                None,
+              ),
+              (0.0, 0.0, 95.0),
+            ),
```
sebastian-peter, line 307: To add to my last comment, this would be an example. Why does it matter whether the heat pump has been running before? Aren't we deciding on that at this point in time anew?

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
```diff
@@ -232,55 +232,209 @@ class HpModelSpec
...
+                  Some(
+                    ThermalStorageState(0L, KilowattHours(500), Kilowatts(0))
+                  ),
+                ),
+                Some(HouseTemperatureUpperBoundaryReached(0L)),
+              ),
+              (0.0, 0.0, 95.0),
+            ),
+
+            // No capacity for flexibility at all
```
sebastian-peter, line 352: ... because both storage and house are at max capacity? Maybe add this to the comment

sebastian-peter (changes requested):
Some questions and comments from my side

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: I don't think so. Might be a copy paste think, since we have TableFor2 / 3 / 4 ...
But the type annotations have no auto complete since for Table one need to provide them as string. 
```
        val testCases =
          Table(
            ("thermalState: ThermalGridState", "lastState: HpState", "expectedValues: (Double, Double, Double)"),
            ```

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: I will summarise our discussion for future reference: This is correct. As we are only testing the FlexOptions here, we can even use the same tick to keep the test case simple. 
The current state depends on the last power, because that is how we distinguish what the reference behaviour will be in this case. For example: House is between lower and target temperature. If the HP was active in the last state, the house was in the heating phase, so our reference behaviour will still be 'active HP'. If it was not active in the last state, the house is cooling, our reference behaviour should be non-active hp until the house reaches the lower temp boundary.

I will add an explanatory comment.

This might be easier to test / need to be changed once we introduced the transitoryData.

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: The current state depends on the last power, because that is how we distinguish what the reference behaviour will be in this case. For example: House is between lower and target temperature. If the HP was active in the last state, the house was in the heating phase, so our reference behaviour will still be 'active HP'. If it was not active in the last state, the house is cooling, our reference behaviour should be non-active hp until the house reaches the lower temp boundary.

docs/readthedocs/models/thermal_grid_model.md
danielfeismann: Simulation will start running but the hp is never used since there is no heat demand. So it's currently kind of mandatory since it's the only heat sink we have...

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
sebastian-peter: Just removed the type specs and it works, let me know what you think

sebastian-peter (approved):
Ok, all good if my proposed change is fine for you too.

src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala
danielfeismann: Yes it works, but if you want to have the types you have to write them manually in Intellij. But I'm fine with this very little and only downside (maybe it's still possible and it's on me).

sebastian-peter: Here, the types are all defined within the `forAll` clause anyway, maybe that suffices?

PR #831, opened 2024-06-16
FixedLoadModelSpec.scala-convert-groovy-to-scala

pierrepetersmeier:
FixedLoadModel convert from groovy to scala.

This PR is related to Issue #646


danielfeismann (changes requested):
This looks already very good. Just some minor remarks for further improvements from my side

src/test/scala/edu/ie3/simona/model/participant/load/FixedLoadModelSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+import org.scalatest.prop.TableDrivenPropertyChecks
+import squants.Power
+import squants.energy.{KilowattHours, Kilowatts, Watts}
+
+class FixedLoadModelSpec
+    extends UnitSpec
+    with LoadInputTestData
+    with TableDrivenPropertyChecks {
+
+  private implicit val tolerance: Power = Watts(1d)
```
danielfeismann, line 29: Maybe not needed any longer.

danielfeismann (changes requested):

src/test/scala/edu/ie3/simona/model/participant/load/FixedLoadModelSpec.scala
pierrepetersmeier: approximate() requires a tolerance as i understand it.

danielfeismann: Yes, thats true. Was still at the stage of should Be ... 

Good catch!

danielfeismann (approved):
Looks very good to me!

src/test/scala/edu/ie3/simona/model/participant/load/FixedLoadModelSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+        (ActivePower(Watts(268.6d)), Watts(268.6)),
+        (EnergyConsumption(KilowattHours(3000d)), Watts(342.24)),
+      )
+
+      forAll(testData) { (reference, expectedPower: Power) =>
+        val relevantData = FixedLoadModel.FixedLoadRelevantData
+
+        var scale = 0.0
+        while (scale <= 2) {
+          val scaledSRated = Kilowatts(
```
danielfeismann, line 121: As Sebastian stated correctly: Using var is usually not what we want, but rather immutable values. We should look into this and try to find such a solution. Maybe this can be done within #886 or as an extra issue / PR

sebastian-peter: Opened up issue here: #919. Could be solved together with the issue that Daniel mentioned or separately

PR #845, opened 2024-07-03
Write ThermalHouse results when em controlled

danielfeismann:
merge #842 first

resolves #844

PR #853, opened 2024-07-17
Fix hp results leading to overheating house and other effects

### LINKED ISSUE #827: Heat pumps heats up houses over upper temperature boundary.
Target temperature: 20
Lower limit: 18
Upperl limit: 22

image

As seen in the picture the house is heated up to almost 23°C when simulating with a cylindrical storage at the house without an EnergyManagementAgent.

image

When simulating without a thermal storage the house is overheated less. 

It can be discussed wether to stop at 20°C or 22°C.

### LINKED ISSUE #844: Write ThermalHouse results when em controlled
At the moment thermal house results are not written out when the heat pump of the house is em controlled

Furthermore the fill_level of cylindrical thermal storages is not calculated correct

danielfeismann:
resolves #827

merge #845 first!


danielfeismann:
Description what caused the error / wrong behaviour

### Heat pump not under control of EmAgent point 1
When the lower temperature limit of the heat house is reached, the heat pump starts to operate, but not immediately when the limit is reached. This is caused by the fact that the maybeThreshold is calculated correctly, but is not added to the additionalActivationTicks.


to be updated

danielfeismann:
Description what caused the error / wrong behaviour #2

### House got overheated again Point 4

Once the house has reached the upper temperature level, the heat from the HP is used to charge the thermal storage. At the next arrival of weather data, the thermal house will be triggered again to determine an updated state. In the current version of `handleInfeed()`, the thermal input into the storage is set to zero to use the heat for the house first. This heats the house again until the upper limit is reached, and so on...

With the changes proposed so far, the hp will now run to heat the house first, then charge the storage, but will continue to run as the house cools down while the storage is being charged, and will therefore have a possible additional heat demand which will then be covered by the hp. Therefore, the demand calculation for the next state needs to be adjusted to only have an additional demand when the inner temperature is below the target temperature.
see `operatesInNextState()`
` demand.hasRequiredDemand || (state.isRunning && demand.hasAdditionalDemand)`

danielfeismann:
Description what caused the error / wrong behaviour part 3

### Point 5 Hp runs longer than expected 
Should be fixed by calculating the actual state within `ThermalGrid.energyDemand()`

### Point 9 ThermalStorageResults have multiple entries
`ThermalGrid.results()` will carry out both, thermalHouse and thermalStorage results whenever one of both got a change of its state. Since the state of the other hasn't changed, it was written out with the tick of the last change of it. This can and should be changed within another PR (Issue needs to be created). As workaround the tick of the result that triggered the output of the results will be used for both.

### Point 10 qDot result still positive when storage is already full 
Should be fixed by calculating the actual state within `ThermalGrid.energyDemand()`

## Still open: 

### Point 8 
When storage is empty the heat pump will not be started immediatly to refill the storage. It only starts, when the thermal house temp drops. (Should be improved)

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -58,24 +58,46 @@
...
+            updatedState._1.innerTemperature < thermalHouse.targetTemperature
+          ) {
+            house
+              .zip(state.houseState)
+              .map { case (house, state) =>
+                house.energyDemand(
+                  tick,
+                  ambientTemperature,
+                  state,
+                )
```
staudtMarius, line 82: Would it be possible to use the above `thermalHouse` and `updatedState` directly here?

staudtMarius (changes requested):
Some small question.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
danielfeismann: Good question but this brought me to another point which is a very good catch! 

I guess, this is not correct. But please double check as well: The part above updates the lastState to the current tick. To determine the energyDemand from that tick on, one needs to use the updatedState, not the (last)state. Or?

staudtMarius: I think you need the updated state if you want to determine the current and future energy demand. If for example the house was heated up before and you use the state that does not include this information, it could lead to the overheating of the house.

Therefore I think you need the current state that includes the current temperature.

danielfeismann: I agree. Guess there might be improvement labeling the states to avoid these kind of failures. I changed to the updatedState already.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -46,36 +46,59 @@ final case class ThermalGrid(
...
     *   Currently applicable state of the thermal grid
     * @return
     *   The total energy demand of the grid
     */
   def energyDemand(
       tick: Long,
-      ambientTemperature: Temperature,
+      // FIXME this is also in state -> refactoring by HiWi
+      lastAmbientTemperature: Temperature,
+      actualAmbientTemperature: Temperature,
```
danielfeismann, line 62: Note from my side: In my opinion both temperatures are necessary here, the ambientTemperature of the last state and the actual one. Since the HpState contains the lastAmbientTemperature this can be done nicer and would be a good HiWi-Job. As well some tests for different cases like falling and rising temperature would be great.

danielfeismann: Transfered this into #916

danielfeismann:
Closed in favour of #931 and #928 and others

PR #860, opened 2024-07-23
Fix hp thermal storage refill

danielfeismann:
resolves point 8 of #827 

merge #853, #825, #918 and #895 first!


danielfeismann:
Closed in favour of #931 and #928 and others

PR #865, opened 2024-07-29
Introducing energy demand for warm water heating

### LINKED ISSUE #856: Enhance thermal house model by heat demand for tap water
With the current version of the thermal house model only heat demand by thermal losses of the building is considered. Since with increasing insulation of houses the share of heat demand for tap water becomes more important our model should be able to include this behavouir as well.

Sources for this can or should be DIN EN 12831-3

danielfeismann:
resolves #856


danielfeismann:
closed for #1358

PR #883, opened 2024-08-05
Provide actual ambient temperature of tick to HpModel when calculate state

### LINKED ISSUE #882: Provide actual ambient temperature of tick to HpModel when calculate state
```
    val (thermalGridState, maybeThreshold) =
      thermalGrid.updateState(
        relevantData.currentTick,
        state.thermalGridState,
        state.ambientTemperature.getOrElse(relevantData.ambientTemperature),
        newThermalPower,
      )
```

imho this should be just `relevantData.ambientTemperature`. Otherwise there could be the situation where updated weather data will not be considered in the very next tick.

danielfeismann:
fix #882


sebastian-peter (changes requested):
I'm not sure if this is what we want. The reason the state's ambient temperature is considered before the "new" temperature from relevant data is the following:

In `handleControlledPowerChange`, we use `calcState` to determine the current state. What we're given is the old state and current relevant data. But to do any calculation at the time of the old state, we also have to use the old ambient temperature - the new one might be different, which leads us to wrong results.

I agree though that the way this is currently handled is quite confusing. So I'd propose taking the following measures:
- Rename `state` to `lastState` in `calcState` and `determineState` as well as in `HpModelSpec` and `ParticipantAgentFundamentals#updateState` and its subclasses
- Correcting the ScalaDoc of `calcState` and `determineState`

danielfeismann:
I still have to check handleControlledPowerChange, so maybe both is right. I'm quite confident that in `ThermalGrid.handleConsumption()` we expect the actual ambientTemperature since we determine the next threshold. When we do so at the moment we get updated weather data, this update is not considered but should so in my opinion. (see PR for thermalGridIT)

sebastian-peter:
@danielfeismann This has been made obsolete by #927, right?

danielfeismann:
Yes, this is fixed by #927

PR #885, opened 2024-08-06
Provide option to directly zip the output files

### LINKED ISSUE #793: Provide option to directly zip the output files
see #781

staudtMarius:
Resolves #793


danielfeismann (commented):
This looks quite fine, however I should have been a bit more precise. Sorry. It would be better, if only the single output files got zipped within rawOutputData. e.g. pv_res.tar.gz.

Could you please adapt this?

src/main/scala/edu/ie3/simona/main/RunSimonaStandalone.scala
```diff
@@ -56,9 +61,38 @@ object RunSimonaStandalone extends RunSimona[SimonaStandaloneSetup] {
...
+              compressFuture.onComplete {
+                case Success(_) =>
+                  FileIOUtils.deleteRecursively(filePath)
+                case Failure(exception) =>
+                  logger.error(
+                    s"Compression of output file to '$archiveName' has failed. Keep raw data.",
+                    exception,
+                  )
+              }
+              Await.ready(compressFuture, 5.minutes)
```
danielfeismann, line 89: Do we / you have any feeling about a proper value here. We had some huge simulations in the past where a single result file had up to 20 Gb. Even this is seldom and one can adapt this value in these cases, we should think about it here.

staudtMarius: We could either increase this time (for example to 1 hour or more) or we could make this cofigurable via the provided config.

danielfeismann: Thats the point. I rather would not go via config. But I'm not sure if 5 minutes are sufficient. Maybe 15 will be a good start? What do you think?

staudtMarius: Do you know how long compressing can take if we have a 20 GB file?

danielfeismann: That was my initial question :) 

I have no idea. Depends of course on the machine, etc... maybe you can find out what would be an good estimation? Otherwise I would go with 15 minutes.

danielfeismann (approved):
LGTM, thanks!

src/main/scala/edu/ie3/simona/main/RunSimonaStandalone.scala
```diff
@@ -56,9 +62,38 @@ object RunSimonaStandalone extends RunSimona[SimonaStandaloneSetup] {
...
+                  logger.error(
+                    s"Compression of output file to '$archiveName' has failed. Keep raw data.",
+                    exception,
+                  )
+              }
+              Await.ready(compressFuture, compressTimeoutDuration)
+            }
+          }
+        }
+
```
sebastian-peter, line 94: Sorry, only looked into this now. Maybe this would've fit better inside `ResultEventListener` while handling `StopTimeout` (i.e. right before the listener stops). This might look pedantic, but I think it's better to keep functionality where it belongs structure-wise. Will provide a suggestion later.

PR #893, opened 2024-08-07
ExtEvDataService returns dummy prices on request

### LINKED ISSUE #892: Handle MobSim requests for current prices
Currently, requesting prices is still supported by simonaAPI and MobSim, but not SIMONA. 
Thus, SIMONA should at least return dummy values (0d) when asked for prices, so that all possible request are considered.

sebastian-peter:
Resolves #892


danielfeismann (approved):
LGTM!

PR #900, opened 2024-08-14
Migration towards pureconfig

### LINKED ISSUE #608: Exchange Config Framework
Since we have had enough issues in the past with our current configuration derivation I think it's time to switch frameworks.

I'm currently considering https://pureconfig.github.io/ as the main option, as it it also reads HOCON files. We should be able to rip out the old framework, add the new one without breaking older configurations that we have used.

Switching to pureconfig also gives us much more flexibility on how we want to use and setup the configuration as well as the option to document the parameters meaning within the classes, allow more flexible converters for specific fields, e.g. DateTime fields and so on.

sebastian-peter:
Resolves #608

This needs some additional work:
- Merge `dev` and handle merge conflicts
- Test whether this works with the configuration files that we currently have


staudtMarius:
Will be resolved with #1169.

PR #902, opened 2024-08-14
Converting all instances of `eval-rst` to the equivalent myst syntax

### LINKED ISSUE #901: Convert all `eval-rst` instances in rtd to myst syntax
From #866 (and before), there's eval-rst instances that can be converted to myst markdown syntax

sebastian-peter:
Resolves #901

A similar thing could probably be done for PSDM as well, btw


danielfeismann (approved):
Very nice that we can get rid of this.

PR #918, opened 2024-08-21
Refactor ThermalEnergyDemand

### LINKED ISSUE #917: Refactor ThermalGridDemand
In my opinion the ThermalGridDemand should 
- [ ] also be able to use for negative values to show some 'production'

Definitions in ScalaDoc state `required` as the absolutely required demand and `possible` for the maximum demand. We named it additionalDemand later, which is a bit misleading. In my point of view we should use it in the meaning of min and max. Thus 
- [ ] `def hasAdditionalDemand` should check if possible > zero not > required
- [ ] Optional: Maybe some functions for `hasRequiredNegativeDemand` (maybe find a better wording) make sense?
- [ ] ... more ... ?

Not sure about:
- [ ] We should throw an exception if possible < required or at least think if the actual solution to set ThermalDemand to (possible, possible) always make sense and is what we expect..

I started but maybe someone would take over this

danielfeismann:
resolves #917

I started here, but more needs to be done. Feel free to take over :)


src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -594,8 +601,16 @@ object ThermalGrid {
...
     ): ThermalEnergyDemand = {
-      if (possible < required)
-        new ThermalEnergyDemand(possible, possible)
+      if (
+        math.abs(possible.toKilowattHours) < math.abs(
+          required.toKilowattHours
+        ) && math.signum(possible.toKilowattHours) <= math.signum(
+          required.toKilowattHours
+        )
+      )
```
danielfeismann, line 610: If possible would be -2 and required 1 this would not throw an exception, or am I wrong?
Lets go with just positive values (at least for now). So it would be great to check first if req or pos is negative plus if possible is smaller than required. In any case we should throw an an InvalidParameterException with a suitable message.

danielfeismann (commented):
Thanks for pushing this further, @marvinheintze! Also a good idea to add the cases for negative demands. I thought a bit on this and I guess it will be easier to go just with positive demands as we at the moment doesn't have any specific use case that requires negative demands. So please remove them again, I'm sorry for the confusion that may arised.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
marvinheintze: Oh yes, you're right. I removed that if statement now and added a new InvalidParameterException for the cases that possible or required are less than zero.

sebastian-peter (approved):
LGTM, thanks!

PR #923, opened 2024-08-24
Fix activation of Hp when not under control of an EM

### LINKED ISSUE #922: Fix activation of Hp when not under control of an EM
#### Point 1
When reaching the lower temp boundary of thermal house the Hp is start working -> 2019-01-18 08:00 (check why activated at the next full hour)

Hp should activated directly when reaching the lower temp boundary and not at the next activation (due to new weather data)

danielfeismann:
resolves #922


danielfeismann:
### Description what caused the error / wrong behaviour

When the lower temperature limit of the heat house is reached, the heat pump starts to operate, but not immediately when the limit is reached. This is caused by the fact that the maybeThreshold is calculated correctly, but is not added to the additionalActivationTicks.

src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
```diff
@@ -261,8 +261,26 @@ trait HpAgentFundamentals
...
+      updatedState.maybeThermalThreshold match {
+        case Some(nextThreshold)
+            if baseStateData.foreseenDataTicks.headOption.exists {
+              case (_, Some(tick)) => nextThreshold.tick < tick
+              case _               => false
+            } =>
+          baseStateData.copy(
+            stateDataStore = updatedStateDataStore,
+            additionalActivationTicks =
+              baseStateData.additionalActivationTicks + nextThreshold.tick,
```
sebastian-peter, line 275: `+` is deprecated, this is not:

```suggestion
              baseStateData.additionalActivationTicks ++ Set(nextThreshold.tick),
```

sebastian-peter (approved):
Thanks, looks good!

sebastian-peter (approved):

PR #925, opened 2024-08-24
Fixed ThermalStorageResults having multiple entries

### LINKED ISSUE #924: Fix ThermalStorageResults having multiple entries
#### Point 9
ThermalStorageResults have multiple entries (2019-01-19 18:00, 2020-01-20 02:00), while the hour following to the double entries is missing.

danielfeismann:
resolves #924


danielfeismann:
ThermalGrid.results() will carry out both, thermalHouse and thermalStorage results whenever one of both got a change of its state. Since the state of the other hasn't changed, it was written out with the tick of the last change of it. This can and should be changed within another PR (Issue needs to be created). As workaround the tick of the result that triggered the output of the results will be used for both.

sebastian-peter (commented):
So now we're setting the result tick to the maximum tick of all results. But wouldn't it make more sense to only send out the results of the entity with the maximum tick? Because the other entity with a smaller result tick clearly has not changed the result.

danielfeismann:
Good idea! Thanks.

sebastian-peter:
@danielfeismann Adapted the code a bit for simplicity. Let me know what you think

danielfeismann:
Looks good to me!

sebastian-peter (approved):

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -399,40 +403,45 @@ final case class ThermalGrid(
...
-      state: ThermalGridState
-  )(implicit startDateTime: ZonedDateTime): Seq[ResultEntity] = {
-    val houseResults = house
+  def results(currentTick: Long, state: ThermalGridState)(implicit
+      startDateTime: ZonedDateTime
+  ): Seq[ResultEntity] = {
+
+    val maybeHouseResult = house
       .zip(state.houseState)
+      .filter { case (_, state) => state.tick == currentTick }
```
danielfeismann, line 422: Looks to me that this rather should be lastTick. Since we yet haven't any ThermalGridIT this comes not directly clear.

' .filter { case (_, state) => state.tick == lastTick}'

PR #933, opened 2024-08-24
Refactoring of Hps createCalcRelevantData to fallback on latest weatherData

### LINKED ISSUE #932: Refactoring of Hps createCalcRelevantData to fallback on latest weatherData if weatherData of currentTick is not available
tba

danielfeismann:
…if weatherData of currentTick is not available

resolves #932


danielfeismann:
Fixed by #956

PR #946, opened 2024-09-02
Enabling EmAgents to handle initialization

### LINKED ISSUE #945: EmAgents should be able to handle initialization
In contrast to regular simulation, during initialization, flex options, flex control and results do not have to be communicated (as they are not calculated).

sebastian-peter:
Resolves #945


danielfeismann (approved):
LGTM

PR #952, opened 2024-09-09
Simplified methods to values in ThermalHouse

### LINKED ISSUE #940: Simplify `ThermalHouse`
I'm not sure if these kinds of one-operation commands need to be placed in their own function:

https://github.com/ie3-institute/simona/blob/51226842952bcea0253c2fb380865b65a0338e33/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala#L225-L229

Especially because this is not understandable at once glance:

https://github.com/ie3-institute/simona/blob/51226842952bcea0253c2fb380865b65a0338e33/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala#L210-L213

But this could be, in my opinion:
```scala
    val temperatureChange = calcInnerTemperatureChange(thermalEnergyChange)
    currentInnerTemperature + temperatureChange
```

Same applies to methods `calcInnerTemperatureChange`, `calcThermalEnergyChange` and `calcThermalEnergyGain`. One could add explanatory commentary if and where needed.

marvinheintze:
Resolves #940 

Changed the methods `calcInnerTemperatureChange`, `calcThermalEnergyChange `and `calcThermalEnergyGain` to values `thermalEnergyLoss`, `thermalEnergyGain`, `temperatureChange`.

`calcInnerTemperatureChange `and `calcNewInnerTemperature `were put together in `val temperatureChange`

`thermalConductanceToEnergy` was changed to `calcThermalEnergyChange`

`def calcThermalEnergyLoss` was removed and is instead calculated in value `thermalEnergyLoss`

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -199,101 +199,19 @@ final case class ThermalHouse(
       currentInnerTemperature: Temperature,
       ambientTemperature: Temperature,
   ): Temperature = {
-    val thermalEnergyChange = calcThermalEnergyChange(
-      calcThermalEnergyGain(thermalPower, duration),
-      calcThermalEnergyLoss(
-        currentInnerTemperature,
-        ambientTemperature,
-        duration,
-      ),
-    )
-    calcNewInnerTemperature(
+    val thermalEnergyGain = thermalPower * duration
+
+    val thermalEnergyLoss = ethLosses.calcThermalEnergyChange(
       currentInnerTemperature,
-      calcInnerTemperatureChange(thermalEnergyChange),
-    )
-  }
-
-  /** Calculate the new inner temperature of the thermal house
-    *
-    * @param oldInnerTemperature
-    *   previous inner temperature
-    * @param temperatureChange
-    *   temperature change
-    * @return
-    *   new inner temperature
-    */
-  def calcNewInnerTemperature(
-      oldInnerTemperature: Temperature,
-      temperatureChange: Temperature,
-  ): Temperature =
-    oldInnerTemperature + temperatureChange
-
-  /** Calculate the temperature change for the thermal house form the thermal
-    * energy change
-    *
-    * @param thermalEnergyChange
-    *   thermal energy change (e.g. through heat pump)
-    * @return
-    *   temperature change
-    */
-  def calcInnerTemperatureChange(
-      thermalEnergyChange: Energy
-  ): Temperature = {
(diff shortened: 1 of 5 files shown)


src/test/scala/edu/ie3/simona/model/thermal/ThermalHouseSpec.scala
```diff
@@ -56,23 +56,19 @@ class ThermalHouseSpec extends UnitSpec with HpInputTestData {
...
+
       val innerTemperatureChange =
-        thermalHouseTest.calcInnerTemperatureChange(thermalEnergyChange)
-      val newInnerTemperature = thermalHouseTest.calcNewInnerTemperature(
-        innerTemperature,
-        innerTemperatureChange,
-      )
+        thermalEnergyChange / thermalHouseTest.ethCapa
+
+      val newInnerTemperature = innerTemperature + innerTemperatureChange
```
sebastian-peter, line 71: This test can just be removed alltogether. The methods that were formerly tested don't exist anymore.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
sebastian-peter, line 1: Looks good so far. Now we've gone a little bit _too_ far in the other direction and a bit more explanation would be nice 😄 The goal would be to understand what is happening at first glance.

My suggestions (open for other/additional suggestions though):
- For `val thermalEnergyLoss`: Add comment above `// thermal energy loss due to the deviation between outside and inside temperature`
- Introduce `val energyChange = thermalEnergyGain - thermalEnergyLoss`

sebastian-peter (changes requested):
Please don't forget adding to changelog as well :)

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: To be fair: Marvin and myself discussed some of these points and I might pushed it a bit to far in the other direction :) We should add a comment and take Sebastians suggestions

marvinheintze: I've added 2 additionial comments, I thought couldn't hurt. You can let me know if they should stay there or if i should remove them.

sebastian-peter (approved):
LGTM

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
sebastian-peter: Looks good, thanks!

PR #956, opened 2024-09-13
Fix expected secondaryData in baseStateData

### LINKED ISSUE #955: Fix expected secondaryData in baseStateData
When we're not em-controlled (in handleActivationAndGoToHandleInformation()) we would expect weatherData for all ticks in baseStateData.receivedSecondaryDataStore. 

https://github.com/ie3-institute/simona/blob/8eab695b3d22afb67671213e0754db5dc70a54f1/src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala#L598-L611

danielfeismann:
resolves #955


sebastian-peter (approved):
Looks good, thanks for fixing this!

PR #960, opened 2024-09-17
TMP: Re-implementation of ParticipantAgent and models

### LINKED ISSUE #959: Re-implementation of ParticipantAgent
This is going to be a rather big task. What we want is:
- More clear and modular design
- Pekko typed
- Built-in handling of states

sebastian-peter:
Resolves #959

This will take some time...

Notes: 
- Retire enabling/disabling capability of `SystemComponent`, was never used
- Retire limitation of reactive power?
- Retire `calculateMissingReactivePowerWithModel`
- Participant replaying primary data is considering operation interval and is able to handle flex power control


sebastian-peter:
All remaining changes are now contained in #1234.

PR #963, opened 2024-09-19
Fixed random failling power flow with em agents

### LINKED ISSUE #962: Fix power flow calculation with em
In random cases, the power flow calculation fails because the power request arrives before the activation.

jo-bao:
Resolves #962


sebastian-peter (approved):
Thanks!

PR #969, opened 2024-09-24
Rewrote ChpModelTest from groovy to scala.

pierrepetersmeier:
Rewrote ChpModelTest from groovy to scala.

resolves: #646


sebastian-peter (commented):
Just marking this as a draft for clarity

PR #971, opened 2024-09-25
convert system participant test to scalatest remove groovy

pierrepetersmeier:
convert system participant test to scalatest remove groovy

resolves: #646


src/test/scala/edu/ie3/simona/model/participant/SystemParticipantSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+      (1.00, Kilovars(0)),
+      (1.03, Kilovars(0)),
+      (1.05, Kilovars(31.22)),
+      (1.07, Kilovars(62.44)),
+      (1.1, Kilovars(62.44)),
+    )
+
+    forAll(testCases) { (adjustedVoltageVal, expectedQ) =>
+      val adjustedVoltage = Each(adjustedVoltageVal)
+      val p = Kilowatts(42)
```
danielfeismann, line 230: Please note, this was originally 100 kW. It is working correct as well at operation point of 42 kW but we should not change the test with this PR.

danielfeismann (changes requested):
Almost there, just a little change

danielfeismann (approved):
LGTM, thanks!

PR #974, opened 2024-09-27
Simplifying integration in QuantityUtil, fixing test

### LINKED ISSUE #973: Simplifying quantity integration in `QuantityUtil`
Implementation should also be a lot faster by making use of SortedMap.

sebastian-peter:
Resolves #973 

Algorithm should now be a bit easier to grasp and run faster due to usage of SortedMap and its specific range methods.


danielfeismann (approved):
Looks good! Thanks for improving!

PR #990, opened 2024-10-10
cylindrical thermal storage test convert groovy tests to scalatest remove groovy

pierrepetersmeier:
Rewrote CylindricalThermalStorageTest Test from groovy to scala

Resolves: #646


src/test/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorageSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+import tech.units.indriya.unit.Units
+
+import java.util.UUID
+
+class CylindricalThermalStorageSpec
+    extends UnitSpec
+    with Matchers
+    with BeforeAndAfterAll {
+
+  final val tolerance: Energy = KilowattHours(1e-10)
```
danielfeismann, line 30: If you change this into an implicit val, you could remove the explicit given tolerances below

src/test/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorageSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+      getQuantity(20, StandardUnits.VOLUME),
+      getQuantity(30, StandardUnits.TEMPERATURE),
+      getQuantity(40, StandardUnits.TEMPERATURE),
+      getQuantity(1.15, StandardUnits.SPECIFIC_HEAT_CAPACITY),
+    )
+  }
+
+  def buildThermalStorage(
+      storageInput: CylindricalStorageInput,
+      volume: Double,
```
danielfeismann, line 49: This is absolutely fine and it was like this in the groovy tests. But maybe we could (and should) use the units here. Same below for `vol2Energy`. This would make it a bit safer to use theses tests because one has not to take care that the prefix of the double is correct.

src/test/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorageSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+        ),
+        (
+          0L,
+          250.0,
+          -10.0,
+          3600L,
+          -42.0,
+          240.0,
+          ThermalStorage.ThermalStorageThreshold.StorageEmpty(4457L),
+        ),
```
danielfeismann, line 184: This one is a repetition (it already was in groovy). Please remove one.

danielfeismann (changes requested):
This looks quite fine, just some few wishes from my side

src/test/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorageSpec.scala
```diff
@@ -0,0 +1,265 @@
...
+import tech.units.indriya.unit.Units
+
+import java.util.UUID
+
+class CylindricalThermalStorageSpec
+    extends UnitSpec
+    with Matchers
+    with BeforeAndAfterAll {
+
+  final implicit val tolerance: Energy = KilowattHours(1e-10)
```
danielfeismann, line 30: This is fine but i don't thing that the tolerance needs to be final here. Final vals can not be overwritten, actually we don't want to overwrite it so it is fine but I guess we normaly simple use implicit val.

src/test/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorageSpec.scala
```diff
@@ -0,0 +1,265 @@
...
+      UUID.randomUUID(),
+      "ThermalStorage",
+      null,
+      getQuantity(100, StandardUnits.VOLUME),
+      getQuantity(20, StandardUnits.VOLUME),
+      getQuantity(30, StandardUnits.TEMPERATURE),
+      getQuantity(40, StandardUnits.TEMPERATURE),
+      getQuantity(1.15, StandardUnits.SPECIFIC_HEAT_CAPACITY),
+    )
+  }
```
danielfeismann, line 45:
```suggestion
  lazy val storageInput = new CylindricalStorageInput(
      UUID.randomUUID(),
      "ThermalStorage",
      null,
      getQuantity(100, StandardUnits.VOLUME),
      getQuantity(20, StandardUnits.VOLUME),
      getQuantity(30, StandardUnits.TEMPERATURE),
      getQuantity(40, StandardUnits.TEMPERATURE),
      getQuantity(1.15, StandardUnits.SPECIFIC_HEAT_CAPACITY),
    )
```

We try to avoid var whereever possible. One option from my point of view here would be using a lazy val. This one get's initialized at the first use and stays imuteable which is what we want.

danielfeismann (changes requested):
Almost there :)

danielfeismann (approved):
LGTM, thanks for including my comments and for the work done here!

PR #993, opened 2024-10-13
Mh/#921 check window duration

### LINKED ISSUE #921: Properly fix CheckWindow duration
With #875, a fix has been introduced that does not actually get to the root cause of the problem. Now we're just fixing a duration in `CheckWindowPassed` that has been calculated wrong from the beginning.

The actual problem lies in `RuntimeNotifier#completing`: 

https://github.com/ie3-institute/simona/blob/c2b41b523ec6834dcaa81fde5ee2e40e322251e6/src/main/scala/edu/ie3/simona/scheduler/RuntimeNotifier.scala#L114-L116

Instead of `lastStartTime`, `lastCheckWindowTime` should be referenced. That should fix the whole issue (please test though). Parts of #875 can then be reverted.

cc @staudtMarius @danielfeismann

marvinheintze:
Resolves #921 
Took out variable `last` in RunTimeEventLogSink 
Changed `lastStartTime` to `lastCheckWindowTime`


danielfeismann (dismissed):
This looks good to me, thanks for fixing this! Maybe @sebastian-peter would like to have a short look here, so I will not merge directly.

Edit: I test the output, is working fine.

sebastian-peter (approved):
Thanks, I just adjusted the display of duration if we pass multiple check windows at once. Now it should be the same as in dev.

@danielfeismann

PR #1009, opened 2024-10-29
Fix filter for thermal result checking for lastTick not for currentTick

### LINKED ISSUE #1008: Fix filter for thermal result checking for lastTick not for currentTick
Looks to me that this rather should be lastTick. Since we yet haven't any ThermalGridIT this comes not directly clear.

' .filter { case (_, state) => state.tick == lastTick}'

_Originally posted by @danielfeismann in https://github.com/ie3-institute/simona/pull/925#discussion_r1820316530_

danielfeismann:
resolves #1008

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
+++ b/src/main/scala/edu/ie3/simona/agent/ValueStore.scala
@@ -22,7 +22,7 @@ import scala.collection.SortedMap
   */
 final case class ValueStore[+D](
     maxTickSpan: Long,
-    store: SortedMap[Long, D] = SortedMap.empty[Long, D],
+    private val store: SortedMap[Long, D] = SortedMap.empty[Long, D],
 ) {
 
   /** Determine the lastly known data tick, if available. Includes the given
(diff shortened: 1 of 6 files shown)


src/main/scala/edu/ie3/simona/agent/ValueStore.scala
```diff
@@ -22,7 +22,7 @@ import scala.collection.SortedMap
   */
 final case class ValueStore[+D](
     maxTickSpan: Long,
-    private val store: SortedMap[Long, D] = SortedMap.empty[Long, D],
+    store: SortedMap[Long, D] = SortedMap.empty[Long, D],
```
sebastian-peter, line 25: I think we don't need this anymore?

sebastian-peter (commented):
Looks good, if this works for you I'd be ready to merge

src/main/scala/edu/ie3/simona/agent/ValueStore.scala
danielfeismann: True, thanks!

sebastian-peter (approved):
Thanks for this concise and effective PR!

PR #1016, opened 2024-11-12
Fix initialisation freezing on empty primary data.

### LINKED ISSUE #981: Error handling for invalid time of primary data
Currently the simulation stalls when time field of primary data is outside of the range of simulation `startDate` and `endDate`.
We should throw an exception in these cases instead rather than leaving the user waiting and wondering / despairing.

The issue can be reproduced by running simona with vn_simona after setting `vn_simona.conf` to:

```
simona.time.startDateTime = "2012-01-01T00:00:00Z"
simona.time.endDateTime = "2012-01-01T02:00:00Z"
```

### LINKED ISSUE #1014: Initialisation process freezes at vn_simona
## Problem

The simulation freezes within initialisation after point when the simulation start date is change to some value not equal to `simona.time.startDateTime = "2011-01-01T00:00:00Z"`

```
16:46:52.668 INFO  e.i.s.e.l.RuntimeEventListener$ - Initializing Agents and Services for Simulation ... 
16:46:52.717 WARN  e.i.s.a.g.GridAgent$$anonfun$edu$ie3$simona$agent$grid$GridAgent$$initializing$1 - Evcs test_evcs_1 has been removed because no ev movements service is present.
16:46:52.717 WARN  e.i.s.a.g.GridAgent$$anonfun$edu$ie3$simona$agent$grid$GridAgent$$initializing$1 - Evcs test_evcs_2 has been removed because no ev movements service is present.
16:46:52.718 WARN  e.i.s.a.g.GridAgent$$anonfun$edu$ie3$simona$agent$grid$GridAgent$$initializing$1 - The following elements have been removed, as the agents are not implemented yet: Set(BmInput)
```

## tasks
- [ ] evaluate further what is happening and what is causing this behaviour 🔍 
- [ ] find solution 💡 
- [ ] fix it 🔨

staudtMarius:
Resolves #981


sebastian-peter:
> One point to discuss, based on the data in `vn_simona`: Asuming I would increase the simulation start time by one hour from '2011-01-01T00:00:00Z' to `2011-01-01T01:00:00Z`, this would cause the exception because time series `8c04e94e-76b0-4369-a55c-f5e1117fb83e` has no entry for 01:00:00. But with our general approach we define values as valid until a new one at a certain tick arrise. In my humble opinion this case would be vaild, while some earlier start date (e.g. 2010-12-31T23:00:00Z') wouldn't.

That's a neat idea, however this sounds more complicated and could rather be handled in a new PR? For now I'm quite happy that there's an error message at all

danielfeismann:
> > One point to discuss, based on the data in `vn_simona`: Asuming I would increase the simulation start time by one hour from '2011-01-01T00:00:00Z' to `2011-01-01T01:00:00Z`, this would cause the exception because time series `8c04e94e-76b0-4369-a55c-f5e1117fb83e` has no entry for 01:00:00. But with our general approach we define values as valid until a new one at a certain tick arrise. In my humble opinion this case would be vaild, while some earlier start date (e.g. 2010-12-31T23:00:00Z') wouldn't.
> 
> That's a neat idea, however this sounds more complicated and could rather be handled in a new PR? For now I'm quite happy that there's an error message at all

Fair point. Yes we should handle that in a new PR

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -131,32 +132,48 @@ final case class PrimaryServiceWorker[V <: Value](
...
         // Note: The whole data set is used here, which might be inefficient depending on the source implementation.
         source.getTimeSeries.getEntries.asScala
-          .filter { timeBasedValue =>
-            val dateTime = timeBasedValue.getTime
-            dateTime.isEqual(simulationStart) || dateTime.isAfter(
-              simulationStart
-            )
-          }
           .map(timeBasedValue => timeBasedValue.getTime.toTick)
+          .filter(_ >= 0L)
```
sebastian-peter, line 142: Nice simplification!

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -131,32 +132,48 @@ final case class PrimaryServiceWorker[V <: Value](
...
-      val initializedStateData =
-        PrimaryServiceInitializedStateData(
-          maybeNextTick,
-          furtherActivationTicks,
-          simulationStart,
-          source,
-        )
-      (initializedStateData, maybeNextTick)
+      (maybeNextTick, furtherActivationTicks) match {
+        case (maybeNextTick @ Some(tick), furtherActivationTicks)
```
sebastian-peter, line 148: You're shadowing the `maybeNextTick` variable unnecessarily here

```suggestion
        case (Some(tick), furtherActivationTicks)
```

sebastian-peter (dismissed):
I fully agree with @danielfeismann that this solution is far from ideal, but at least we display an error message now.

We should work on a more flexible solution that just takes the last value before simulation start if no value for tick 0 is present, as Daniel suggested. 

One could also talk about sending out some sort of `0 kW` data if the time series starts _after_ simulation start, however, this could lead to strange agent behavior that is difficult to detect. It's probably better to require the user to explicitly set the value for tick 0.

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
sebastian-peter: Just committed this myself because it's my only suggestion here...

sebastian-peter (approved):
Thanks a lot, @staudtMarius, for fixing this!

sebastian-peter (approved):

PR #1028, opened 2024-11-18
Removing pmd

### LINKED ISSUE #1027: Remove pmd
pmd does not have any rules for Scala, but does just duplicate checks, which is also handled by sonarqube

sebastian-peter:
Resolves #1027


PhilippSchmelter (approved):
Looks good to me :)

PR #1038, opened 2024-11-20
Considering primary data that start before simulation.

### LINKED ISSUE #1034: Consider PrimaryData that start before simulation
One point to discuss, based on the data in `vn_simona`: Asuming I would increase the simulation start time by one hour from '2011-01-01T00:00:00Z' to `2011-01-01T01:00:00Z`, this would cause the exception because time series `8c04e94e-76b0-4369-a55c-f5e1117fb83e` has no entry for 01:00:00. But with our general approach we define values as valid until a new one at a certain tick arrise. In my humble opinion this case would be vaild, while some earlier start date (e.g. 2010-12-31T23:00:00Z') wouldn't.

_Originally posted by @danielfeismann in https://github.com/ie3-institute/simona/issues/1016#issuecomment-2473095811_

staudtMarius:
Resolves #1034


sebastian-peter (commented):
Thanks for putting in the effort. I've been looking through the code and kept thinking that we should just integrate this into `getTimeKeysAfter` and make it the PSDM's job.... That'd make it less complicated here.

staudtMarius:
I will create an PSDM issue for this.

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -270,14 +264,52 @@ object PrimaryServiceWorker extends SimonaService {
...
+                simulationStart,
+                valueClass,
+                source,
+              )
+
+            Success(initializedStateData, maybeNextTick)
+
+          case (_, Some(value)) =>
+            /* We have data before, but not after the start of the simulation */
+            log.debug(
```
sebastian-peter: This should be a `log.warn` imho

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -270,14 +264,52 @@ object PrimaryServiceWorker extends SimonaService {
...
+            val initializedStateData =
+              PrimaryServiceInitializedStateData(
+                maybeNextTick,
+                furtherActivationTicks,
+                simulationStart,
+                valueClass,
+                source,
+              )
+
+            Success(initializedStateData, maybeNextTick)
```
sebastian-peter: We should start at tick 0 then, right? 

```suggestion
            Success(initializedStateData, Some(0L))
```

(This also makes more sense when looking at the case before this one, where we fail because we only have data _after_ simulation start.)

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
```diff
@@ -362,25 +382,35 @@ object PrimaryServiceWorker extends SimonaService {
...
-      baseStateData.copy(
-        maybeNextActivationTick = maybeNextActivationTick,
-        activationTicks = remainderActivationTicks,
-      ),
-      maybeNextActivationTick,
-    )
+    val ticks = baseStateData.activationTicks
+
+    baseStateData.maybeNextActivationTick match {
+      case Some(nextTick) if nextTick > currentTick =>
```
sebastian-peter: This does not happen at all, I think. The previous code was fine the way it was. Or do you have any counter-example?

sebastian-peter (changes requested):
Some points from my side, looks mostly ready

src/main/scala/edu/ie3/simona/service/primary/PrimaryServiceWorker.scala
staudtMarius: There is a test, where we request data for tick 200 an receive a completion for tick 900. Therefore, I added this code. Otherwise we would not be scheduled for tick 900.

If you have a better solution for the problem, please let me know.

sebastian-peter: I think the test is erroneous. Here, we are told by the service that we want to get activated at 900:

https://github.com/ie3-institute/simona/blob/8883013c5fc2561e1cda34f72b29115fb9b532ae/src/test/scala/edu/ie3/simona/service/primary/PrimaryServiceWorkerSpec.scala#L400-L407

Right after that, we activate the service at 200:

https://github.com/ie3-institute/simona/blob/8883013c5fc2561e1cda34f72b29115fb9b532ae/src/test/scala/edu/ie3/simona/service/primary/PrimaryServiceWorkerSpec.scala#L412

This should not happen in reality.

staudtMarius: I find it interesting that the actual tick value is not checked. Maybe we should add some code to check the value of the activation message. Or is this already handled elsewhere.

sebastian-peter: This is naturally handled by the scheduler, who only activates actors according to the tick that they scheduled themselves for. I'm not against adding a check here, though, if you think that one is needed.

staudtMarius: After checking the code again, I think we don't need a check here.

sebastian-peter (approved):
LGTM, thanks for being patient with my requests!

PR #1043, opened 2024-11-22
Fix implausible test cases of HpModelSpec

### LINKED ISSUE #1042: Fix implausible test cases of HpModelSpec
https://github.com/ie3-institute/simona/blob/741a2b2b5a7c54b7e59c1a9c71931f1eedb02e33/src/test/scala/edu/ie3/simona/model/participant/HpModelSpec.scala#L121-L134

if the Hp is running in the lastState, the thermalState need to be set up with the qDot of the Hp not only with some temperature (at least if no storage or any other heat sink is available).

danielfeismann:
resolves #1042


danielfeismann:
Closed for #1044

PR #1047, opened 2024-11-22
Correcting wrong use of "wall clock time"

### LINKED ISSUE #727: Correct wrong use of term "wall clock time"
At some places in the code and ScalaDoc, "wall clock time" is used where "simulation time" is meant.

For an explanation, one can consult the source below (or many other sources) or just ask me.

```
R. M. Fujimoto, Parallel and Distribution Simulation Systems. New York: John Wiley & Sons, Inc., 1999.
```

sebastian-peter:
Resolves #727


danielfeismann (approved):
LGTM, thanks for the fix!

PR #1053, opened 2024-11-24
Refactoring to only use 'lastHpState' and 'relevantData' for 'Thermal…

### LINKED ISSUE #916: Refactor: Use last ambientTemperature from lastState
this should build upon #853 

within `ThermalGrid.energyDemand` I introduced the last ambientTemperature very quick and dirty. This can and should be refactored. The ambientTemperature of the last state is also part of the last HpState and can be taken from there. That would make the method more comprehensive. This may lead to further adaptions...

see:

              Note from my side: In my opinion both temperatures are necessary here, the ambientTemperature of the last state and the actual one. Since the HpState contains the lastAmbientTemperature this can be done nicer and would be a good HiWi-Job. As well some tests for different cases like falling and rising temperature would be great.

_Originally posted by @danielfeismann in https://github.com/ie3-institute/simona/pull/853#discussion_r1724463737_

danielfeismann:
…Grid' calculations

resolves #916


sebastian-peter (approved):
Looks good!

PR #1095, opened 2025-01-08
Converted `ExtSimAdapter` to typed.

### LINKED ISSUE #1094: Convert ExtSimAdapter to typed

staudtMarius:
Resolves #1094


sebastian-peter (changes requested):
Maybe you're right and we should wait for Scala 3 here. It's also a bit annoying that we can't create the behavior with type parameter `Request`, which would then also be possible.

staudtMarius:
> Maybe you're right and we should wait for Scala 3 here. It's also a bit annoying that we can't create the behavior with type parameter `Request`, which would then also be possible.

But we could use a custom message adapter, like we did with the external data services.

src/main/scala/edu/ie3/simona/api/ExtSimAdapter.scala
```diff
@@ -6,36 +6,32 @@
...
 
 import scala.jdk.OptionConverters._
 
 object ExtSimAdapter {
 
-  def props(scheduler: ActorRef): Props =
-    Props(
-      new ExtSimAdapter(scheduler)
-    )
+  type Request = Create | Stop | Activation | ControlResponseMessageFromExt
```
sebastian-peter: Hmm, thinking if it made sense to still have a sealed trait (let's call it `ExtSimRequest` here) that is enhanced by the "actual" ExtSimAdapter messages (`Create` and `Stop`), so that the type def would become:

```scala
  type Request = ExtSimRequest | Activation | ControlResponseMessageFromExt
```

This way, it would become clear at one glance which types are exclusive to the actor, and which ones are external.

sebastian-peter (changes requested):

src/main/scala/edu/ie3/simona/api/ExtSimAdapter.scala
staudtMarius: Good idea. I will adapt this part.

sebastian-peter (approved):

PR #1098, opened 2025-01-09
Throw exception if the slack node is not directly conected to a transformer.

### LINKED ISSUE #525: Validate slack node location
When having a slack node not located directly on a transformator node the calculation of the current in the line from slack is not done correct. One possible solution seems to have the slack directly located on a trasnformer node.

grafik

- [ ] Further elaborate this
- [ ] Validate the location of slack at a transformer node through validation utils

staudtMarius:
Resolves #525


staudtMarius:
!test

danielfeismann (approved):
Thanks for handling this error

PR #1105, opened 2025-01-14
Upgrade `scala2` to `scala3`.

### LINKED ISSUE #53: Switch to scala 3 + java 17

staudtMarius:
Resolves #53 

Merge #1374 first.


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Updated
+- Upgraded `scala2` to `scala3` [#53](https://github.com/ie3-institute/simona/issues/53)
+
+## [4.0.0] - 2025-05-09
```
danielfeismann: this line should be removed

build.gradle
```diff
@@ -32,6 +32,18 @@ ext {
...
+  scala3compilerOptions = [
+    "-source:3.7-migration",
+    "-print-tasty",
+    "-feature",
+    "-language:implicitConversions",
+    "-language:existentials",
+    "-Wunused:all",
+    "-Xmax-inlines:128"
+    //"-explain",
+    //"-explain-types"
```
danielfeismann: do we need this?

src/main/scala/edu/ie3/simona/agent/grid/congestion/CongestedComponents.scala
```diff
@@ -108,7 +108,7 @@ object CongestedComponents {
           val vMag =
             vNom * nodeRes(transformer.lvNodeUuid).getvMag.toSquants.toEach
 
-          val power = sqrt(3.0) * res.getiBMag.toSquants * vMag
+          val power = res.getiBMag.toSquants * sqrt(3.0) * vMag
```
danielfeismann: It triggers my inner monk that the sqrt can't stand at the start of the formula. Just a personal opinion, so you could leave it as it is, but I would place the sqrt 3 at the end in these cases.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -67,12 +67,12 @@ class ThermalGridIT
...
     with AnyWordSpecLike
     with should.Matchers
+    with QuantityMatchers
     with EmInputTestData
     with MockitoSugar
     with DefaultTestData
     with TestSpawnerTyped {
-  private implicit val quantityTolerance: Double = 1e-10d
-  protected val temperatureTolerance: Double = 0.01
+  protected given temperatureTolerance: Double = 0.01
```
danielfeismann: This is quite dangerous, I've already fallen into that trap myself. Now this tolerance applies to all doubles, e.g. also the power, qDot, etc. This is not what we want.

danielfeismann: But if you have an approach to get rid of the explicit tempTolerance, I would be very grateful.

danielfeismann (changes requested):
Some points from my side

build.gradle
staudtMarius: I used them to get a more detailed explanation why the code didn't compile. I will remove them.

Some of the other compiler options can be removed after we adapted the whole code.

src/main/scala/edu/ie3/simona/agent/grid/congestion/CongestedComponents.scala
staudtMarius: A added a new method in `edu.ie3.util.scala.quantities.QuantityUtil`. With this new extension, we can keep the code as it was before.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
staudtMarius: If we want remove the tolerance, we have to change the default tolerance value in the `equalWithTolerance` method. I found a way, where the temperature tolerance will only be used for checking temperature values.

danielfeismann (approved):
LGTM

PR #1113, opened 2025-01-20
Sending flex options even if not em-controlled

### LINKED ISSUE #1112: EM should output flex option results even if it has no parent

sebastian-peter:
Resolves #1112


danielfeismann (approved):
LGTM :)

PR #1108, opened 2025-01-15
Prepare changes for next psdm release

danielfeismann:
resolves #1107

PR #1118, opened 2025-01-22
Use ValidationUtils for ThermalGrids

### LINKED ISSUE #1117: Use ValidationUtils also on AssetInput not only for JointGridContainer
The only place where we use ValidationUtils to check for valid input data is here 

https://github.com/ie3-institute/simona/blob/dev/src/main/scala/edu/ie3/simona/io/grid/GridProvider.scala#L52-L53

Since there are assets that are not included in the JointGridContainer we don't use ValidationUtils to check those.
This effects
- thermal houses
- thermal storages
- maybe ems?
- maybe more?

Since there are ValidationUtils at least for ThermalUnits, we should use them in SIMONA.

danielfeismann:
resolves #1117 
at least for thermalGrids and their assets.

requires: 
- https://github.com/ie3-institute/PowerSystemDataModel/pull/1217 
- #1108


src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
```diff
@@ -456,6 +457,7 @@ trait HpAgentFundamentals
         s"Unable to initialize heat pump agent '${inputModel.electricalInputModel.getUuid}' without thermal grid model."
       )
     case WithHeatInputContainer(_, thermalGrid) =>
+      ValidationUtils.check(thermalGrid)
```
staudtMarius, line 460: Is there a reason, for doing the validation here? For the `JointGridContainer` we do the check right after reading it (see: `GridProvider`).

staudtMarius (changes requested):
I have a small question regarding this PR.

src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
danielfeismann: Right thats the place where we check normally everything. At the moment i don't see a (simple) way to access the complete thermalGrid there. But maybe I oversaw something? If you have a simple approach checking this at GridProvider, feel free. 

Nonetheless maybe a comment in GridProvider would be good, that the validation of thermalGrid is done in HpAgentFundamentals...

staudtMarius: You could do it in the map method, where we map the thermal grid to its bus. You could first valudate the grid and than map it.

Another option would be to add one of the following lines before the mapping: 
- `.map(ValidationUtils.check)`
- `.map(ValidationUtils.check(_))`

src/main/scala/edu/ie3/simona/agent/participant/hp/HpAgentFundamentals.scala
```diff
@@ -11,6 +11,7 @@ import edu.ie3.datamodel.models.result.system.{
   HpResult,
   SystemParticipantResult,
 }
+import edu.ie3.datamodel.utils.validation.ValidationUtils
 import edu.ie3.simona.agent.ValueStore
```
staudtMarius, line 15: This import can be removed.

staudtMarius (changes requested):
The import in the `HpAgentFundamentals` is no longer necessary.

staudtMarius (approved):
LGTM

PR #1104, opened 2025-01-13
Added support for topologies without transformers and slack grids with multiple nodes

staudtMarius:
Resolves #1099

PR #1125, opened 2025-01-27
Enhance exception message in case of InvalidGridException

### LINKED ISSUE #1124: Enhance exception message in case of InvalidGridException
`The grid contains no basic branch elements (lines or transformers).`

should give a bit more context

danielfeismann:
resolves #1124


src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
```diff
@@ -407,7 +407,7 @@ object GridModel {
       noLines && noTransformers2w && noTransformers3w && (noOfNodes > noOfSlackNodes)
     )
       throw new InvalidGridException(
-        "The grid contains no basic branch elements (lines or transformers)."
+        f"The grid with subnet number ${gridModel.subnetNo} contains no basic branch elements (lines or transformers)."
```
sebastian-peter, line 410: You could go one step further and add this: `Only grids that only consist of slack nodes are allowed to not contain lines and transformers` (or similar)

 (I had to deal with this not long ago and it took some time to understand the situation)

danielfeismann: Thanks, main issue was to get the subnet that raises the issue. Will adapt this, asking myself if it is longer valid with https://github.com/ie3-institute/simona/pull/1104

pierrepetersmeier:
!test

src/main/scala/edu/ie3/simona/model/grid/GridModel.scala
sebastian-peter: That's a good question, but since you're only adapting the error message to describe current reality, this is fine for now imho

sebastian-peter (approved):

PR #1132, opened 2025-02-04
Enhance EmAggregate of SelfOpt to cope with other targetLimits

### LINKED ISSUE #1131: Enhance EmAggregate of SelfOpt to cope with other targetLimits
merge PowerSelfOpt with same functionalities for some powerLimit that shouldn't be exceeded if possible...

danielfeismann:
resolves #1131


danielfeismann:
!test

src/test/scala/edu/ie3/simona/model/em/EmAggregatePowerOptSpec.scala
```diff
@@ -0,0 +1,688 @@
...
+      )
+
+      actualResult shouldBe (
+        Kilowatts(-3.0),
+        Kilowatts(-7.0),
+        Kilowatts(4.0)
+      )
+    }
+
+    "pick (reference power (positive) if inside the power limit" in {
```
staudtMarius, line 480: I think this bracket can be removed.

src/main/scala/edu/ie3/simona/model/em/EmAggregatePowerOpt.scala
```diff
@@ -12,29 +12,38 @@ import edu.ie3.simona.ontology.messages.flex.MinMaxFlexibilityMessage.ProvideMin
...
+  * @param targetPower
+  *   power limit that should be not be exceeded
   * @param curtailRegenerative
   *   Whether to include positive flexibility of PV/WEC in reference sum
   *   calculation
   */
-final case class EmAggregateSelfOpt(curtailRegenerative: Boolean)
-    extends EmAggregateFlex {
+final case class EmAggregatePowerOpt(
+    targetPower: Power = zeroKW,
```
staudtMarius, line 29: Maybe we can find more suitable name for `targetPower`, because for values, except the default value, it behaves more like an interval. Also, in the tests it is sometimes called a `limit` and in the `CHANGELOG` it is called `targetLimit`.

staudtMarius (changes requested):
This looks good. Just some small points.

src/main/scala/edu/ie3/simona/model/em/EmAggregatePowerOpt.scala
danielfeismann: I see your point. I startet with limit but it isn't really a limit since we allow to exceed if it's infeaseable. So target was and is still my prefered choice. But I would be happy, if you have a good suggestion. Nevertheless I will harmonise the other limits that might be in there...

staudtMarius: We could name it something like `targetRange` or `targetPowerRange`, to express that we trying to find a value, that is inside of a set range.

danielfeismann: Yes, but the value itself is one target value, not a range. Therefor I'll opt for targetPower.

staudtMarius: But isn't it `±targetPower`?

danielfeismann: yes, but from users perspective range could be interpreted to give two values, which is not supported.

staudtMarius: Should we name it `targetPowerAbs` or at least document, that this value is seen as an absolute value?

danielfeismann: Good idea!

staudtMarius (approved):
LGTM

PR #1164, opened 2025-02-10
Implementing Version Check Step in GHA Pipeline

### LINKED ISSUE #1067: Implement version check in GHA
It should work similarly to how it was done with Jenkins.

Normally, dev version should be higher than the one in main. After merging dev, versions in main and dev should be the same (shortly). When merging back changes of the release from main to dev, the snapshot version should increase again.

main | dev
0.5.0 | 0.6.0 <- Before release
0.6.0 | 0.6.0 <- After release PR has been merged
0.6.0 | 0.7.0 <- After changes from main have been merged back to dev

This means: 
- For PRs merging onto main (in a release branch from dev), we're allowed to increase the version in main to the version in dev
- For PRS merging onto dev (from feature branches), the dev version should stay the same (we're not allowed to change the version)
- There is an exception to the last rule though: If main and dev versions are equal, we're allowed to increase the dev version from our feature branch
- When the pipeline runs for dev or main itself, no version check should occur.

So a control structure could be the following:
```
if PR:
  if base == dev:
    if dev version == branch version:
      ok
    else:
      if main version == dev version & branch version > dev version:
        ok
      else:
        fail
  else: # base == main
    if branch version > main version:
      ok
    else:
      fail
else: # no PR
  ok
```

PhilippSchmelter:
Resolves #1067

This pull request includes several changes to improve the GitHub Actions pipeline and add version checking scripts. The most important changes include modifications to the CI workflow configuration, the addition of new scripts for branch type detection and version checking, and updates to the `CHANGELOG.md` file.

### CI Workflow Improvements:
* [`.github/workflows/ci.yml`](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR48-R60): Added the `Version Check` step to the CI workflow, which runs the `run-all.sh` script when a pull request is created.

### New Scripts:
* [`scripts/branch_type.sh`](diffhunk://#diff-33fe978e5913b775b3c2a2e0adc584f3e0aff5cdca36fd19acab4904bb2a4020R1-R39): Added a new script to determine the branch type based on the branch name pattern.
* [`scripts/get_versions.sh`](diffhunk://#diff-d8d9c2770513b4850061c25220bf6a0727b62daf9fa2c4c9a5600886aba24a0cR1-R38): Added a script to fetch the current version of the pull request and the versions from the `dev` and `main` branches.
* [`scripts/run-all.sh`](diffhunk://#diff-01046df752455fc45cf94e973dbb217681cb055822642e0a3bc51166146187e8R1-R13): Added a script to execute version fetching and checking scripts.
* [`scripts/version_check.sh`](diffhunk://#diff-bac2665431f40a52c4e46a07da5ffea08ca73d50a04bb5db54a700c1feb00b5dR1-R76): Added a script to check if the pull request version is appropriate based on the target branch (`dev` or `main`).

### Documentation Update:
* [`CHANGELOG.md`](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR60): Updated to include the addition of the `Version Check` to the GitHub Actions pipeline.


sebastian-peter (changes requested):
Thanks, this looks quite promising already!

A few wishes from my side:
- Please put the script inside some `scripts/version_check.sh`
- I realized that my issue description was not well thought-through. I enhanced it a bit, which should give a bit more guidance. But please think it through yourself before implementing, I may have made mistakes.
- Please don't forget the changelog addition!

PhilippSchmelter:
> Thanks, this looks quite promising already!
> 
> A few wishes from my side:
> 
> * Please put the script inside some `scripts/version_check.sh` (if that's possible in a convenient way with all the variables etc.)
> * I realized that my issue description was not well thought-through. I enhanced it a bit, which should give a bit more guidance. But please think it through yourself before implementing, I may have made mistakes.
> * Please don't forget the changelog addition!

Thanks for helping me out with this one! I admit the first Issue description was somehow confusing for me. The updated version helped me understand what is needed :)

I think what i've done this time should meet your requirements :D
At this point i need to transform the GHA step into .sh script. But I tested every case (except the bump rule and I know the bump logic is not right. For this to work we need to commit and push, because the changes made in the pipeline are just local.)

Looking forward to some feedback!

.github/workflows/ci.yml
```diff
@@ -49,6 +49,39 @@ jobs:
           
           ./gradlew checkBranchName -PbranchName="$branch_name"
 
+      - name: Get Versions
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          REPO_URL="https://github.com/ie3-institute/simona.git"
```
sebastian-peter, line 55: Maybe there's a [way to get the URL programatically](https://stackoverflow.com/a/65461907)? That way, you could just copy the code as-is to other repositories later. You might not need the TOKEN that the linked solution uses.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,39 @@ jobs:
...
+          rm -rf dev-branch
+          
+          #MAIN-Version
+          git clone --depth 1 --branch main "$REPO_URL" main-branch
+          cd main-branch
+          MAIN_VERSION=$(./gradlew -q currentVersion)
+          cd ..
+          echo "MAIN_VERSION=$MAIN_VERSION"
+          echo "MAIN_VERSION=$MAIN_VERSION" >> $GITHUB_ENV
+          rm -rf main-branch
```
sebastian-peter, line 77: I thought about using bash scripts here so that all the cloning could be done there as well (which would be difficult in gradle). So maybe you could only save REPO_URL in env and handle the rest in the script? That way you could also only add a single `Version Check` job

scripts/branch_type.sh
sebastian-peter, line 1: Looks like a nice script, but is it used anywhere?

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+# Version Checking Logic
+if [ "$BASE_BRANCH" = "dev" ]; then
+  echo "PR into dev => applying dev rules"
+  if [ "$DEV_VERSION" = "$PR_VERSION" ]; then
+    echo "OK: dev version == PR version"
+    exit 0
+  else
+    if [ "$MAIN_VERSION" = "$DEV_VERSION" ]; then
+      if semver_gt "$PR_VERSION" "$DEV_VERSION"; then
+        echo "Bump dev"
```
sebastian-peter, line 56: A more descriptive print text would be nice :) Maybe something like "OK: Increasing working version in dev from $DEV_VERSION to $PR_VERSION", or something like that

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+        exit 0
+      else
+        echo "FAIL: dev == main, but PR version is NOT greater than dev."
```
sebastian-peter, line 59:
```suggestion
        echo "FAIL: Release and working version are $MAIN_VERSION, but PR is not increasing the working version in dev"
```

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+        exit 1
+      fi
+    else
+      echo "FAIL: dev != PR version, and dev != main."
```
sebastian-peter, line 63: "dev != main" should be the regular case, but "dev != PR" is not allowed then. So maybe the message should explain that regular PRs must not update the working version, or something like that.

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+      exit 1
+    fi
+  fi
+
+elif [ "$BASE_BRANCH" = "main" ]; then
+  if semver_gt "$PR_VERSION" "$MAIN_VERSION"; then
+    echo "OK: PR version is greater than main version"
```
sebastian-peter, line 70:
```suggestion
    echo "OK: Release is increasing the version from $MAIN_VERSION to $PR_VERSION"
```

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+  else
+    return 1
+  fi
+}
+
+# Version Checking Logic
+if [ "$BASE_BRANCH" = "dev" ]; then
+  echo "PR into dev => applying dev rules"
+  if [ "$DEV_VERSION" = "$PR_VERSION" ]; then
+    echo "OK: dev version == PR version"
```
sebastian-peter, line 51:
```suggestion
    echo "OK: Working version in dev stays $DEV_VERSION"
```

scripts/version_check.sh
```diff
@@ -0,0 +1,80 @@
...
+      exit 1
+    fi
+  fi
+
+elif [ "$BASE_BRANCH" = "main" ]; then
+  if semver_gt "$PR_VERSION" "$MAIN_VERSION"; then
+    echo "OK: PR version is greater than main version"
+    exit 0
+  else
+    echo "FAIL: PR version is NOT greater than main version"
```
sebastian-peter, line 73: Please adapt accordingly by including the actual versions like above

sebastian-peter (changes requested):
Some more comments :)

.github/workflows/ci.yml
PhilippSchmelter: Done

.github/workflows/ci.yml
PhilippSchmelter: After our meeting I tried to implement it in a single step by using source. I couldnt find a clean way of implementation, so i went back to two steps. This way we can use GITHUB_ENV. Get Versions is still in a .sh format

.github/workflows/ci.yml
```diff
@@ -45,9 +45,19 @@ jobs:
             branch_name="${branch_name#refs/heads/}"
           fi
           
+          export branch_name
+          
           echo "branch_name=$branch_name" >> $GITHUB_ENV
           
-          ./gradlew checkBranchName -PbranchName="$branch_name"
+          ./gradlew checkBranchName -PbranchName="$branch_name" --warning-mode=none
```
sebastian-peter, line 52: Just to be sure: Why did you add the `--warning-mode=none` flag?

scripts/get_versions.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "${BRANCH_NAME^^}_VERSION=$BRANCH_VERSION" >> "$GITHUB_ENV"
+
+    rm -rf "$DIR_NAME"
+}
+
+
+get_branch_version "dev"
+get_branch_version "main"
+
+echo "Get Versions: OK!"
```
sebastian-peter, line 38: Please don't forget the final line break :)

scripts/run-all.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail
+
+rm -f versions.env
+
+chmod +x scripts/get_versions.sh
+chmod +x scripts/version_check.sh
```
sebastian-peter, line 7: Wondering why this is necessary, as you don't have to do set the other `.sh` as executable. I think it's possible to commit scripts as executable with git. Check [this](https://stackoverflow.com/a/40979016) if it doesn't work with Intellij natively

scripts/run-all.sh
sebastian-peter, line 1: Maybe a more descriptive name could be found? `run-version-check.sh`?

scripts/get_versions.sh
```diff
@@ -0,0 +1,38 @@
...
+
+REPO_URL=$(git config --get remote.origin.url)
+export REPO_URL
+echo "REPO_URL=$REPO_URL" >> $GITHUB_ENV
+
+echo "Fetching current version of PR..."
+PR_VERSION=$(./gradlew -q currentVersion)
+echo "PR_VERSION=$PR_VERSION"
+echo "export PR_VERSION=$PR_VERSION" >> versions.env
+echo "PR_VERSION=$PR_VERSION" >> "$GITHUB_ENV"
```
sebastian-peter, line 14: Question: So `export` is used for the scripts to work, and the github env is used for subsequent steps in the pipeline to receive the variables?

sebastian-peter (changes requested):
Looks a lot cleaner, especially considering the averse conditions of having to use shell scripts :)

scripts/get_versions.sh
PhilippSchmelter: Yes! Absolutely right: Export is necessary so we can use the variables in the script itself and Line 14 is to keep the variables saved in GitHub_ENV :)

.github/workflows/ci.yml
sebastian-peter: Thanks, makes sense!

.github/workflows/ci.yml
```diff
@@ -45,9 +45,19 @@ jobs:
             branch_name="${branch_name#refs/heads/}"
           fi
           
+          export branch_name
+          
           echo "branch_name=$branch_name" >> $GITHUB_ENV
```
sebastian-peter, line 50: Could you capitalize the env variable name, as you did with all others?

scripts/branch_type.sh
```diff
@@ -0,0 +1,39 @@
...
+elif [[ "$branch_name" =~ $pattern_feature ]]; then
+  branch_type="feature"
+else
+  echo "Error:'$branch_name' does not match any pattern."
+  exit 1
+fi
+
+echo "========================="
+echo "Branch type: $branch_type"
+echo "branch_type=$branch_type" >> "$GITHUB_ENV"
```
sebastian-peter, line 38: Same here...

sebastian-peter (changes requested):
One more small thing...

PR #1166, opened 2025-02-11
Fix determination of qDotIntoGrid in HpModel.calcState() in case heat…

### LINKED ISSUE #1165: Fix determination of qDotIntoGrid in `HpModel.calcState()` in case heatStorage should feed the thermal grid
https://github.com/ie3-institute/simona/blob/dev/src/main/scala/edu/ie3/simona/model/participant/HpModel.scala#L236

`lastStateStorageQDot == zeroKW && (demandWrapper.houseDemand.hasRequiredDemand || demandWrapper.heatStorageDemand.hasRequiredDemand)`

The second part of the OR condition does not really make sense, instead we should check if the storage isn't empty.

danielfeismann:
…Storage should feed the thermal grid

resolves #1165


src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
```diff
@@ -223,17 +224,17 @@ final case class HpModel(
       isRunning: Boolean,
       demandWrapper: ThermalDemandWrapper,
   ): HpState = {
-    val lastStateStorageQDot = lastState.thermalGridState.storageState
-      .map(_.qDot)
-      .getOrElse(zeroKW)
+    val lastStateStorage = lastState.thermalGridState.storageState
+      .getOrElse(ThermalStorageState(-1, zeroKWh, zeroKW))
```
danielfeismann, line 228: @reviewer: Do you prefer having this or a check for storage and some exception?

src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
```diff
@@ -222,24 +230,32 @@ final case class HpModel(
...
         (pRated, pThermal, pThermal)
-      else if (lastStateStorageQDot < zeroKW)
-        (zeroKW, zeroKW, lastStateStorageQDot * (-1))
+      // If the house has req. demand and storage isn't empty, we can heat the house from storage.
       else if (
-        lastStateStorageQDot == zeroKW && (demandWrapper.houseDemand.hasRequiredDemand || demandWrapper.heatStorageDemand.hasRequiredDemand)
+        currentEnergyOfThermalStorage > zeroKWh && demandWrapper.houseDemand.hasRequiredDemand
+      ) {
+        (zeroKW, zeroKW, pThermalOfStorage)
+        // Edge case when em controlled: If the house was heated last state by Hp and surplus energy is gone now,
```
sebastian-peter, line 253: What's "surplus energy"?
Also, you mean "has not reached target temp yet", or did it actually just try at the last tick?

src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
```diff
@@ -222,24 +230,32 @@ final case class HpModel(
...
+        // but house didn't reach target temperature yet. House can be heated from storage, if this one is not empty.
+      } else if (
+        currentEnergyOfThermalStorage > zeroKWh && demandWrapper.houseDemand.hasAdditionalDemand && qDotLastHouseState > zeroKW
```
sebastian-peter, line 256: Whats additionalDemand for the house again and why doesn't requiredDemand suffice, since the comment line above mentions target temperature, and the documentation of thermal demand says that "boundary or target temperature" is reached when the _required_ demand is fulfilled?

sebastian-peter (changes requested):
Some questions, as usual, from my side about the HP functionality...

src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
danielfeismann: Changed surplus to `and setPower is below turnOn condition now,`. Within lastState we heated the house and we're still below targetTemp, then we would continue heating from storage even if the new Em request to reduce electric power, right?

src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
danielfeismann: Good catch. At this point, additionalDemand is still when "boundary or target temperature" are reached, depending on the setPower the Em demands. But this will change to target temperature with #1183

sebastian-peter (approved):
Ok then, LGTM

PR #1168, opened 2025-02-12
Check for runningHp when handling infeed to thermalGrid

### LINKED ISSUE #1167: Handle edge cases of thermal grid
these are

- hp is em-controlled. Was heating the house with surplus energy from pv. House is still below targetTemperature when a weather change cancels the energy surplus of the pv plant. In this case we would like to continue heating from storage if there is one.

danielfeismann:
resolves #1167


sebastian-peter (approved):
LGTM

PR #1169, opened 2025-02-12
Switching to `pureconfig`.

### LINKED ISSUE #608: Exchange Config Framework

staudtMarius:
Resolves #608


staudtMarius:
!test

build.gradle
```diff
@@ -128,8 +127,9 @@ dependencies {
 
   /* config */
   implementation 'com.typesafe:config:1.4.3'
-  implementation "com.github.carueda:tscfg_2.13:$tscfgVersion"
   implementation "com.github.scopt:scopt_${scalaVersion}:4.1.0" // cmd args parser
+  implementation "com.github.pureconfig:pureconfig_${scalaVersion}:${pureConfigVersion}"
+  implementation "com.github.pureconfig:pureconfig-pekko_${scalaVersion}:${pureConfigVersion}"
```
sebastian-peter, line 132: `pureconfig-pekko` seems to be not used. You could also remove the `pureConfigVersion` variable then

src/main/scala/edu/ie3/simona/config/ArgsParser.scala
```diff
@@ -30,7 +30,7 @@ object ArgsParser extends LazyLogging {
       nodePort: Option[String] = None,
       seedAddress: Option[String] = None,
       useLocalWorker: Option[Boolean] = None,
-      tArgs: Map[String, String] = Map.empty,
+      tArgs: Map[String, String] = Map.empty, // TODO: can be removed?
```
sebastian-peter, line 33: Think so, but separate issue and then remove this todo?

src/main/scala/edu/ie3/simona/config/ArgsParser.scala
```diff
@@ -177,6 +177,7 @@ object ArgsParser extends LazyLogging {
           ReflectionTools
             .resolveClassNameToCompanion(listenerElem.fullClassPath) match {
             case Some(listener: SimonaListenerCompanion) =>
+              // TODO: Double check if we should have an Optional here by design
```
sebastian-peter, line 180: So you mean rather `resolveClassNameToCompanion` returning the `Try` and working with that here?

src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
```diff
@@ -6,14 +6,85 @@
...
+  def render(options: ConfigRenderOptions): String =
+    SimonaConfig.render(this, options)
+}
+
 object SimonaConfig {
+  // pure config start
+  implicit def productHint[T]: ProductHint[T] =
+    ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
+
+  // TODO: replace with finite duration
```
sebastian-peter, line 31: Is this done in another PR, or shoudl we create an issue for this?

src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
```diff
@@ -6,14 +6,85 @@
...
+              f"CannotRead => Can not read config source} \n"
+            case ConvertFailure(reason, _, path) =>
+              f"ConvertFailure => Path: $path, Description: ${reason.description} \n"
+            case ThrowableFailure(throwable, origin) =>
+              f"ThrowableFailure => ${throwable.getMessage}, Origin: $origin \n"
+            case failure =>
+              f"Unknown failure type => ${failure.toString} \n"
+          }
+          .mkString("\n")
+        throw new RuntimeException(
```
sebastian-peter, line 65: Please make this a `CriticalFailureException`

src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
```diff
@@ -2902,9 +2978,9 @@ object SimonaConfig {
     }
 
     final case class Time(
-        endDateTime: java.lang.String,
-        schedulerReadyCheckWindow: scala.Option[scala.Int],
-        startDateTime: java.lang.String,
+        endDateTime: String = "2011-05-01T01:00:00Z",
+        schedulerReadyCheckWindow: Option[Int] = None,
+        startDateTime: String = "2011-05-01T00:00:00Z",
```
sebastian-peter, line 2983: Don't think that there should be default simulation start and end, as this is very individual to the specific simulation. But I realize this is how it was done beforehand, so maybe we should do this in a separate issue

sebastian-peter (changes requested):
Looks all good to me, thanks a lot @t-ober @marvinheintze @staudtMarius! Just some minor questions and remarks from my side

src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
staudtMarius: There is an PR @marvinheintze worked on.

src/main/scala/edu/ie3/simona/config/ArgsParser.scala
staudtMarius: This comment was from @t-ober. From my perspective, we can leave this as it is now.

src/main/scala/edu/ie3/simona/config/ArgsParser.scala
staudtMarius: See #1178

src/main/scala/edu/ie3/simona/config/SimonaConfig.scala
staudtMarius: see #1179

sebastian-peter: Thanks!

sebastian-peter (approved):
Ok, great! Thanks!

PR #1171, opened 2025-02-12
Cleaning up `SimonaConfig`.

### LINKED ISSUE #1170: Clean up `SimonaConfig`
After switching to `pueconfig`, we can remove the generated methods.

staudtMarius:
Resolves #1170


sebastian-peter (approved):
@staudtMarius I just replaced all instances of `List()` with `List.empty` for the reason explained [here](https://stackoverflow.com/questions/14814593/why-use-collection-emptyt-instead-of-new-collectiont/14815359#14815359)

PR #1177, opened 2025-02-13
changed pvInput values and corresponding expected results

### LINKED ISSUE #1144: Refactor pvInput in PvInputTestData to realistic values
The current implementation of pvInput in PvInputTestData uses some unrealistic values. Perhaps we can fix this and refactor the tests that would then fail.

At first glance these are

albedo = 1 => 0.2
etaConv = 10% (Quantities.getQuantity(10, StandardUnits.EFFICIENCY)) => ~ 90%
elevation_angle = 100  (Quantities.getQuantity(100, StandardUnits.SOLAR_ELEVATION_ANGLE)) => 45°
kg = 12 => 0.9
kT =11 => 1.0

marvinheintze:
Resolves #1144

Changed the pvInput values und the corresponding equalWithTolerance values in EmAgentIT


sebastian-peter (commented):
LGTM, but I'll have @danielfeismann have the final say

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -285,9 +285,11 @@ class EmAgentIT
             emResult.getInputModel shouldBe emInput.getUuid
             emResult.getTime shouldBe 0L.toDateTime
             emResult.getP should equalWithTolerance(
-              -0.00034885012.asMegaWatt
+              -0.03283549601246891.asMegaWatt
```
danielfeismann: These values are correct, but not really suited to the test situation. I also noticed that we scaled the pv above (almost certainly because it didn't produce enough with the old values). I changed that and fixed the values for the first tick. I'll commit my changes, could you please adjust the others? Please also change the comment accordingly.

danielfeismann: Maybe you need to debug to get the new values from PvModel (best option debug in PvModel.calcOutput())

danielfeismann (changes requested):
Thanks for tackling this issue! I found some artefacts which I'll fix directly (scaling of PV).

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
danielfeismann: Maybe a bit more context: The Pv has 10 kW, before my change it was scaled by factor 2. So its max power would be arround 20 kW. The weather with `WattsPerSquareMeter(540d)` and `WattsPerSquareMeter(400d)` may be a bit too strong sunshine (Pv does produce arround 38kW according to the emResult). So I would propose to reduce the irradiance of the weather to match (roughly) the old pv production. If there are any questions, please don't hesitate to ask :)

marvinheintze: Ah thank you for the context, that is very helpful! I'll adjust the other values.

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -55,14 +55,14 @@ import squants.motion.MetersPerSecond
 import squants.thermal.Celsius
 
 import java.time.ZonedDateTime
+import scala.concurrent.duration.{FiniteDuration, SECONDS}
```
danielfeismann, line 58: This import is no longer used, thus it can be removed.

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -150,7 +150,7 @@ class EmAgentIT
               pvInput,
               PvRuntimeConfig(
                 calculateMissingReactivePowerWithModel = true,
-                scaling = 2d,
+                scaling = 1d,
```
danielfeismann, line 153: I missed this when I changed it. Since default parameter for scaling is 1, this line could be removed. 

Same for scaling and initicalSoc for Storage below. Could you please remove them as well? Thanks!

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -436,7 +442,7 @@ class EmAgentIT
...
```
danielfeismann, line 445: same here for the scaling

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -285,18 +285,20 @@ class EmAgentIT
...
 
         scheduler.expectMessage(Completion(emAgentActivation, Some(7200)))
 
         /* TICK 7200
          LOAD: 0.269 kW (unchanged)
-         PV:  -3.651 kW
-         STORAGE: SOC 63.3 %
-         -> charge with 3.382 kW
+         PV:  -3.791 kW
+         STORAGE: SOC 63.3 % fixme: reminder check this at the end
```
danielfeismann, line 300: SoC is still at 63.3 % => 5000 W * 0.95 (eta of converter) = 4750 W, 2h of operation => 9500 Wh / 15000 Wh capacity = 63,3%

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -559,10 +565,10 @@ class EmAgentIT
 
         /* TICK 0
          LOAD: 0.269 kW
-         PV:  -5.617 kW
+         PV:  -5.842 kW
          Heat pump: off, can be turned on or stay off
          -> set point ~3.5 kW (bigger than 50 % rated apparent power): turned on
-         -> remaining -0.499 kW
+         -> remaining -0.072 kW
```
danielfeismann, line 571: this should be -0.723 kW, right?

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -348,10 +352,10 @@ class EmAgentIT
 
         /* TICK 14400
          LOAD: 0.269 kW (unchanged)
-         PV:  -0.066 kW
+         PV:  -0.069 kW
          STORAGE: SOC 100 %
-         -> charge with -0.202956 kW
-         -> remaining 0 kW
+         -> charge with 0 kW
```
danielfeismann, line 357: We do charge here with the remaing power. So this is 0.2 kW

danielfeismann (approved):
LGTM, found one little last thing and fixed it directly. Thanks for improving our test cases!

PR #1209, opened 2025-02-19
Operating `PvModel` with data validity duration based on the next tick

### LINKED ISSUE #1208: Operate `PvModel` with data validity duration based on the next, not the last tick
I'd like to adapt the old PvModel to the new logic used with the new model in a separate issue

sebastian-peter:
Resolves #1208

The data validty duration (duration in which the current tick is valid) is now based on the interval "current tick -> next tick" (as it should), and not anymore based on "last tick -> current tick".


sebastian-peter:
Closed in favor of solution for #1212

PR #1224, opened 2025-02-24
Setting a higher testing timefactor

### LINKED ISSUE #1222: Fix test timeouts for all tests
Randomly, tests keep failing like the following, probably because in CI the timeout is exceeded because we run many tests in parallel.

```
edu.ie3.simona.sim.SimonaSimSpec > SimonaSim should indicate success to the starter and stop when receiving SimulationEnded from TimeAdvancer FAILED

    java.lang.AssertionError at SimonaSimSpec.scala:94
```

Set a higher test timeout per default.

sebastian-peter:
Resolves #1222


staudtMarius (approved):
LGTM

PR #1232, opened 2025-02-25
Fix handling of state in `ParticipantModel` and `ParticipantModelShell`

### LINKED ISSUE #1228: Fix handling of states in `ParticipantModelShell`
The determination of the current state is not correct, currently:

https://github.com/ie3-institute/simona/blob/5f867a46d0077fccd38fb92f6d0ccaaf0497b8c1/src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala#L456-L463

If the state for the current tick has already been calculated, it is not returned, but the initial state.

Check if this can be tested.

sebastian-peter:
Resolves #1228

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
+++ b/src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
@@ -430,18 +430,22 @@ final case class ParticipantModelShell[
   private def determineCurrentState(tick: Long): S = {
     // new state is only calculated if there's an old state and an operating point
     val state = _state
-      .zip(_operatingPoint)
-      .map { case (st, op) =>
-        if (st.tick < tick)
+      .map { st =>
+        if (st.tick < tick) {
+          // If the state is old, an operating point needs
+          // to be present to determine the curren state
           model.determineState(
             st,
-            op,
+            operatingPoint,
             tick,
             tick.toDateTime(simulationStartDate),
           )
-        else
+        } else {
+          // The state is up-to-date, no need to update
           st
+        }
       }
+      // No state present, create an initial one
       .getOrElse(model.initialState(tick, tick.toDateTime(simulationStartDate)))
 
     if (state.tick != tick)
--- a/src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
+++ b/src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
@@ -24,7 +24,6 @@ import edu.ie3.simona.exceptions.CriticalFailureException
 import edu.ie3.simona.model.SystemComponent
 import edu.ie3.simona.model.em.EmTools
 import edu.ie3.simona.model.participant2.ParticipantModel.{
-  ModelInput,
   ModelState,
   OperatingPoint,
   OperationChangeIndicator,
@@ -67,9 +66,6 @@ import scala.reflect.ClassTag
   *   The date and time at which simulation started.
   * @param _state
   *   The most recent model state, if one has been calculated already.
-  * @param _input
-  *   The most recent model input used for calculation, if it has been received
-  *   already.
   * @param _flexOptions
   *   The most recent flex options, if they have been calculated already.
   * @param _lastOperatingPoint
@@ -94,7 +90,6 @@ final case class ParticipantModelShell[
     private val operationInterval: OperationInterval,
     private val simulationStartDate: ZonedDateTime,
     private val _state: Option[S] = None,
-    private val _input: Option[ModelInput] = None,
     private val _flexOptions: Option[ProvideFlexOptions] = None,
     private val _lastOperatingPoint: Option[OP] = None,
     private val _operatingPoint: Option[OP] = None,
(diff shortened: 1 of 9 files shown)


src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
```diff
@@ -454,13 +430,23 @@ final case class ParticipantModelShell[
...
-      .flatMap { case (st, op) =>
-        Option.when(st.tick < tick) {
-          model.determineState(st, op, modelInput)
+      .map { st =>
+        if (st.tick < tick) {
+          // If the state is old, an operating point needs
+          // to be present to determine the curren state
+          model.determineState(
+            st,
+            operatingPoint,
```
staudtMarius, line 439: Is there a reason, you no longer use `.zip(_operatingPoint)`?

src/main/scala/edu/ie3/simona/model/participant2/PrimaryDataParticipantModel.scala
```diff
@@ -62,22 +61,27 @@ final case class PrimaryDataParticipantModel[PD <: PrimaryData: ClassTag](
...
-  override val initialState: ModelInput => PrimaryDataState[PD] = { input =>
-    val primaryData = getPrimaryData(input.receivedData)
-    PrimaryDataState(
-      primaryData,
-      input.currentTick,
-    )
+  override val initialState: (Long, ZonedDateTime) => PrimaryDataState[PD] = {
+    (tick, _) =>
+      PrimaryDataState(
+        primaryDataExtra.zero,
```
staudtMarius, line 67: Why did you change the primary data?

src/main/scala/edu/ie3/simona/model/participant2/PvModel.scala
```diff
@@ -74,25 +74,30 @@ class PvModel private (
...
-  override val initialState: ModelInput => PvState = { input =>
-    val weatherData = getWeatherData(input.receivedData)
-
-    PvState(
-      input.currentTick,
-      input.currentSimulationTime,
-      weatherData.diffIrr,
-      weatherData.dirIrr,
-    )
+  override val initialState: (Long, ZonedDateTime) => PvState = {
```
staudtMarius, line 77: The brace is unnecessary.

src/main/scala/edu/ie3/simona/model/participant2/WecModel.scala
```diff
@@ -67,24 +66,28 @@ class WecModel private (
     with ParticipantSimpleFlexibility[WecState]
     with LazyLogging {
 
-  override val initialState: ModelInput => WecState = { input =>
-    val weatherData = getWeatherData(input.receivedData)
+  override val initialState: (Long, ZonedDateTime) => WecState = { (tick, _) =>
```
staudtMarius, line 69: The brace is unnecessary.

src/main/scala/edu/ie3/simona/model/participant2/WecModel.scala
```diff
@@ -67,24 +66,28 @@ class WecModel private (
...
     WecState(
-      input.currentTick,
-      weatherData.windVel,
-      weatherData.temp,
+      tick,
+      MetersPerSecond(0d),
+      Celsius(0d),
```
staudtMarius, line 73: This could be added to the default quantities.

src/test/scala/edu/ie3/simona/agent/participant2/MockParticipantModel.scala
```diff
@@ -52,23 +55,41 @@ class MockParticipantModel(
       MockState,
     ] {
 
-  override val initialState: ModelInput => MockState = { input =>
-    val maybeAdditionalPower = input.receivedData.collectFirst {
-      case data: MockSecondaryData =>
-        data.additionalP
-    }
-
+  override val initialState: (Long, ZonedDateTime) => MockState = { (tick, _) =>
```
staudtMarius, line 58: The brace is unnecessary.

src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
sebastian-peter: Yes, zip only gives you a `Some` if both values are present. That was not correct though for the beginning of the simulation

src/main/scala/edu/ie3/simona/model/participant2/PrimaryDataParticipantModel.scala
sebastian-peter: Because now, a state is determined in two steps (for some complicated reasons) and for the initial state, we don't supply input data anymore.

src/main/scala/edu/ie3/simona/model/participant2/PvModel.scala
sebastian-peter: `val` functions always look a bit more complicated to me without braces, but okay, let's try without

src/main/scala/edu/ie3/simona/model/participant2/WecModel.scala
sebastian-peter: It could, but we're currently only using it one time, so I think it's not that useful

danielfeismann (approved):
Fine for me, but let's check Marius opinion as well.

src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelShell.scala
staudtMarius: Ok.

src/main/scala/edu/ie3/simona/model/participant2/WecModel.scala
staudtMarius: Then we should keep this way.

sebastian-peter:
@staudtMarius @danielfeismann Thanks both of you!

PR #1235, opened 2025-02-26
Added psdm `LoadProfileSource`s.

### LINKED ISSUE #951: Use the PSDM LoadProfileSource for reading in load profiles

staudtMarius:
Resolves #951


src/main/scala/edu/ie3/simona/model/participant2/load/RandomLoadModel.scala
```diff
@@ -0,0 +1,79 @@
...
+    *
+    * It is said in 'Kays - Agent-based simulation environment for improving the
+    * planning of distribution grids', that the Generalized Extreme Value
+    * distribution's parameters are sampled from input data, that is normalized
+    * to 1,000 kWh annual energy consumption. However, due to inaccuracies in
+    * random data reproduction, the sampled values will lead to an average
+    * annual energy consumption of approx. this value. It has been found by
+    * 1,000 evaluations of the year 2019.
+    */
+  private val profileReferenceEnergy = KilowattHours(716.5416966513656)
```
sebastian-peter, line 44: This also only works for the examplary random profile (same point as for the standard load model).

sebastian-peter (changes requested):
Thanks for this implementation draft! I like a lot of the ideas that went into this. However, I'm afraid there's also several flaws that require fixing.
- RandomLoadProfile won't work like this, as you wouldn't want the same power value for all LoadAgents at some given tick.
- `ParticipantAgentInit` does not know how to register for `ServiceType.LoadProfileService`
- We didn't even notice this because the test grid vn_simona does not even contain a single load that is _not_ configured as fixed. Maybe you could add one or two standard profile loads.

staudtMarius:
Since the changes to the participants leads to errors while merging, I created a new brnach: #1357

PR #1236, opened 2025-02-27
Introducing HpModel for the new ParticipantAgent

### LINKED ISSUE #1155: Replace `HpModel` with its new implementation

danielfeismann:
resolves #1155

merge first
- #1252 
- #1183
- #1277
- #1159


src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala
```diff
@@ -0,0 +1,493 @@
...
+    * @param thermalDemands
+    *   The actual thermal demands of the thermal grid elements (house,
+    *   storage).
+    */
+  final case class HpState(
+      override val tick: Long,
+      ambientTemperature: Temperature,
+      thermalGridState: ThermalGridState,
+      lastHpOperatingPoint: HpOperatingPoint,
+      lastStateAmbientTemperature: Temperature,
```
sebastian-peter, line 465: `lastStateAmbientTemperature` should not be needed anymore. In the process of calculating the current state, you should still have acceess to the `ambientTemperature` of the previous state. After calculating the current state, you shouldn't need the previous ambient temp. anymore, as far as I can tell.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -48,39 +54,47 @@ final case class ThermalGrid(
...
     *   The total energy demand of the house and the storage and an updated
     *   [[ThermalGridState]].
     */
   def energyDemandAndUpdatedState(
-      relevantData: HpRelevantData,
-      lastHpState: HpState,
+      tick: Long,
+      state: HpState,
+      operatingPoint: HpOperatingPoint,
   ): (ThermalDemandWrapper, ThermalGridState) = {
```
sebastian-peter, line 71: I got confused looking through the code and was trying to pinpoint the source of my confusion, and I think it's this.

I think energy demand and current state calculation should be separated. For calculating the current state, you should only need the last state and operating point since then. Energy demand on the other hand is something that should lead to the determiantion of the operating point (or flex options), and should take the current state only as input (plus current weather data, but that's besides the point).

src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala
```diff
@@ -0,0 +1,493 @@
...
+    val (newActivePowerHp, qDotIntoGrid) = nextOperatingPoint(state, None)
+
+    /* Push thermal energy to the thermal grid and get its updated state in return */
+    val (updateState, maybeThreshold) = thermalGrid.updateState(
+      state.tick,
+      state,
+      newActivePowerHp > zeroKW,
+      qDotIntoGrid,
+      state.thermalDemands,
+    )
```
sebastian-peter, line 349: Hmm, this also doesn't seem like I intended this. When we're determining the operating point, the state of the grid should already be determined: The parameter `state` is already the current state.

sebastian-peter (changes requested):
Thanks for taking the time to adapt to the new model structure! My first round of questions, let me know if we should discuss more in-depth.

src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala
danielfeismann: I don't agree. You're right in principle, but we would be forcing this concept into ThermalHouse. Maybe it's worth it, maybe not. But I think it should be done in a separate PR. In ThermalHouse.updateState() we actually need the lastAmbientTemp to determine the current state. And the ambientTemp to determine the threshold.

If we split this into two methods we could do what you're suggesting.

src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala
danielfeismann: I guess, we could adapt this. We only need to get back the information how the qDotHp is used from the thermalGrid (for the house, for the storage, so split?) and the next threshold.

But like this, it seems to me more compact. Other thing, is that I now would like to find a better naming for updateState() since there is now also updatedThermalGridState(), I also tried to combine both, but this need a bit more refactoring, I guess.

danielfeismann: Writing this i got some idea :)

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
```diff
@@ -0,0 +1,499 @@
...
+              )
+              minPower should approximate(Kilowatts(expectedMinPower))
+              maxPower should approximate(Kilowatts(expectedMaxPower))
+          }
+
+      }
+    }
+
+    "Handle controlled power change" in {}
+  }
```
danielfeismann, line 498: Ok, just realised that I skipped some tests. Will work on that shortly

src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala
danielfeismann: this would be solved by #1329

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -692,55 +565,101 @@ final case class ThermalGrid(
...
-              ThermalHouseState(tick, innerTemperature, thermalInfeed),
-            ) =>
+    val currentOpThermals = currentOperatingPoint.thermalOps
+
+    val lastOpThermals = lastOperatingPoint match {
+      case Some(op) => op.thermalOps
+      case None     =>
+        // we need some thermals that are different from zero for the first result
+        ThermalOpWrapper(Kilowatts(-42), Kilowatts(-42), Kilowatts(-42))
+    }
```
sebastian-peter, line 593: This could be solved by making lastOpThermals optional (you then need the further adaptions below):

```suggestion
    val lastOpThermals = lastOperatingPoint.map(_.thermalOps)
```

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -692,55 +565,101 @@ final case class ThermalGrid(
...
-            s"Result handling for storage type '${heatStorage.getClass.getSimpleName}' not supported."
+            s"Result handling for storage type '${storage.getClass.getSimpleName}' not supported."
           )
+        )
+    }
+
+    val maybeHouseResult = {
+      (
+        house,
+        currentOpThermals.qDotHouse != lastOpThermals.qDotHouse,
```
sebastian-peter, line 637:
```suggestion
        lastOpThermals.forall(_.qDotHouse != currentOpThermals.qDotHouse),
```

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -692,55 +565,101 @@ final case class ThermalGrid(
...
+        case (Some(house: ThermalHouse), _, false) =>
+          createThermalHouseResult(house)
+        case _ => None
+      }
+    }
+
+    val maybeStorageResult = {
+      (
+        heatStorage,
+        currentOpThermals.qDotHeatStorage != lastOpThermals.qDotHeatStorage,
```
sebastian-peter, line 652:
```suggestion
        lastOpThermals.forall(
          _.qDotHeatStorage != currentOpThermals.qDotHeatStorage
        ),
```

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -701,37 +576,8 @@ class ThermalGridIT
...
-                  energy,
-                ) =>
-              inputModel shouldBe typicalThermalStorage.getUuid
-              time shouldBe 28000.toDateTime
-              qDot should equalWithTolerance(0.asMegaWatt)
-              energy should equalWithTolerance(0.asMegaWattHour)
-          }
-        }
+      // We don't expect any message for house or storage
+      resultListener.expectNoMessage()
```
sebastian-peter, line 580: Same here

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -992,69 +815,64 @@ class ThermalGridIT
       emInitSchedule.tick shouldBe INIT_SIM_TICK
       val emAgentActivation = emInitSchedule.actor
 
+      // We don't expect any message for house or storage
```
sebastian-peter, line 818: Comment doesn't fit here

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
```diff
@@ -218,26 +227,23 @@ class ThermalGridWithHouseAndStorageSpec
...
       "take energy from storage, if there is actual consumption" in {
-        val relevantData = HpRelevantData(
-          0L,
-          testGridAmbientTemperature,
-        )
-        val initialGridState = ThermalGrid.startingState(thermalGrid)
-        val initialLoading = KilowattHours(200d)
-        val gridState = initialGridState.copy(storageState =
-          initialGridState.storageState.map(storageState =>
+        val initialLoading = KilowattHours(10d)
```
sebastian-peter, line 230: Why did you change the initial stored energy?

sebastian-peter (changes requested):
Some minor points I came across

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -462,10 +437,16 @@ class EmAgentIT
         scheduler.expectMessage(Completion(lockActivation))
 
         /* INIT */
-
         emAgentActivation ! Activation(INIT_SIM_TICK)
 
-        primaryServiceProxy.receiveMessages(2) should contain allOf (
+        primaryServiceProxy.receiveMessages(
+          3,
+          FiniteDuration(60, SECONDS),
```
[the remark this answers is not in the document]
danielfeismann: Good catch!

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
danielfeismann: Wanted to keep the storage being empty as reason for the next threshold.

sebastian-peter (approved):
Unironical big achievement 🎉

Issue #1030, opened 2024-11-18
Migrate gradle from Groovy to Kotlin


sebastian-peter:
https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html

PR #1243, opened 2025-03-03
Gradle groovy to kotlin

imoverclocked:
@sebastian-peter I saw #1030 and thought I'd take a stab at the top-level build.gradle file. I have two commits, the first one just renames the files so that the second commit is a little more auditable. ie: I highly recommend viewing the second commit on its own.

While I haven't done exhaustive testing, it does compile and print out a list of tasks when I `./gradlew tasks`

There is probably a lot of low-hanging fruit to be gained in (say) IntelliJ but I'll leave that to whomever is interested in it.


imoverclocked:
No comments in two weeks, moving on.

PR #1247, opened 2025-03-05
Use new tick retrieval methods from PSDM.

### LINKED ISSUE #920: Use the new tick retrieval methods from PSDM
With https://github.com/ie3-institute/PowerSystemDataModel/pull/1059, new methods have been introduced retrieving future activation times for primary data (time series) and weather data. Use these methods in SIMONA during setup. 

This is only possible once the changes have been released, beforehand it's possible to use the snapshot version of PSDM.

staudtMarius:
Resolves #920


sebastian-peter (approved):
LGTM thanks!

PR #1252, opened 2025-03-07
Use input container when init participant models

danielfeismann:


src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelInit.scala
```diff
@@ -13,6 +13,10 @@ import edu.ie3.simona.agent.participant.data.Data.{
   PrimaryData,
   PrimaryDataExtra,
 }
+import edu.ie3.simona.agent.participant.statedata.ParticipantStateData.{
+  InputModelContainer,
+  WithHeatInputContainer,
```
staudtMarius, line 18: I think this import is unnecessary at the moment.

src/test/scala/edu/ie3/simona/test/common/input/EmInputTestData.scala
```diff
@@ -77,7 +81,7 @@ trait EmInputTestData
     Quantities.getQuantity(0.95, PU),
   )
 
-  protected val householdStorageInput = new StorageInput(
+  protected val storageInput = new StorageInput(
```
staudtMarius, line 84: Why did you change this name? Also, if you rename this value, you should probably rename the id too.

staudtMarius (changes requested):
LGTM. Just two comments from my side.

src/main/scala/edu/ie3/simona/model/participant2/ParticipantModelInit.scala
danielfeismann: Yes, thats true. Good catch

src/test/scala/edu/ie3/simona/test/common/input/EmInputTestData.scala
```diff
@@ -24,6 +24,10 @@ import edu.ie3.datamodel.models.input.thermal.{
 }
 import edu.ie3.datamodel.models.input.{EmInput, OperatorInput}
 import edu.ie3.datamodel.models.{OperationTime, StandardUnits}
+import edu.ie3.simona.agent.participant.statedata.ParticipantStateData.{
+  SimpleInputContainer,
+  WithHeatInputContainer,
```
staudtMarius, line 29: This import can be removed too.

src/test/scala/edu/ie3/simona/test/common/input/EmInputTestData.scala
```diff
@@ -77,9 +81,9 @@ trait EmInputTestData
     Quantities.getQuantity(0.95, PU),
   )
 
-  protected val householdStorageInput = new StorageInput(
+  protected val storageInput = new StorageInput(
```
staudtMarius, line 84: Maybe we should also rename the type input.

staudtMarius (changes requested):
Two more small adaptions needed in `EmInputTestData`.

staudtMarius (approved):
LGTM.

staudtMarius (changes requested):
LGTM.

staudtMarius (approved):
LGTM.

PR #1257, opened 2025-03-10
Use Math.floor when calcNextEventTick for charging entities

### LINKED ISSUE #1256: Use Math.floor when calcNextEventTick for charging entities
to always be on the safe side of rounding and not cut energy that overshoot the available storage volume.

https://github.com/ie3-institute/simona/blob/9c80601f216a451751554da5f81dab7b6655d282/src/main/scala/edu/ie3/simona/model/participant2/ChargingHelper.scala#L75-L102

danielfeismann:
resolves #1256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated to changes of PSDM release v6.0.0 [#1107](https://github.com/ie3-institute/simona/issues/1107)
 - Refactor external simulation setup [#1136](https://github.com/ie3-institute/simona/issues/1136)
 - Use new tick retrieval methods from PSDM [#920](https://github.com/ie3-institute/simona/issues/920)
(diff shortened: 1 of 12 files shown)


sebastian-peter (changes requested):
Thanks for the discussion and the initiative.

Found some more examples that we should adapt at the same time:

https://github.com/ie3-institute/simona/blob/9c80601f216a451751554da5f81dab7b6655d282/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala#L108-L109

https://github.com/ie3-institute/simona/blob/9c80601f216a451751554da5f81dab7b6655d282/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala#L115-L116

https://github.com/ie3-institute/simona/blob/9c80601f216a451751554da5f81dab7b6655d282/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala#L339-L342

sebastian-peter (approved):
Sorry, didn't realize that many tests needed to be adapted. Thanks though

sebastian-peter (approved):

PR #1274, opened 2025-03-14
Fix bug in `ExtDataSupport.adapter`.

### LINKED ISSUE #1273: Fixing adapter for external data services.
Due to the way the scheduling of exernal data services works, the external data service adapter will be given to the `ExtSimAdapter`. The `ExtSimAdapter` will send the service a `ScheduleServiceActivation` with leads to an exception, since `ScheduleServiceActivation` does not comply with the expected type `DataMessageFromExt`.

staudtMarius:
Resolves #1273


danielfeismann (approved):
LGTM, should and could we secure this by some test? If not, feel free to merge this

sebastian-peter (approved):
LGTM, test not needed imho

PR #1277, opened 2025-03-14
Fixing initialization scheduling by using ScheduleLocks

### LINKED ISSUE #1272: Fix initialization by using ScheduleLocks
Initialization might currently fail due to race conditions, which occur when participants want to schedule themselves for initialization but it's currently not assured that the scheduling happens before simulation starts.

sebastian-peter:
Resolves #1272


danielfeismann (approved):
LGTM, thanks for also enhancing the documenation!

PR #1279, opened 2025-03-14
Removing `ChpModel`

### LINKED ISSUE #1156: Remove `ChpModel` implementation
Some more extensive adaptations might have to be done here

sebastian-peter:
Resolves #1156


danielfeismann (approved):
Goodbye, may the time come when we'll see each other again.

PR #1303, opened 2025-03-24
Fixed pekko config not written as output

### LINKED ISSUE #1302: Fix pekko config output
Currently, the pekko config is not written ín the output config.

staudtMarius:
Resolves #1302


src/main/scala/edu/ie3/simona/sim/setup/SetupHelper.scala
```diff
@@ -225,13 +226,16 @@ trait SetupHelper extends LazyLogging {
...
     * [[SimonaConfig]], otherwise an exception is thrown
     *
+    * @param typeSafeConfig
+    *   All configuration parameters
     * @param simonaConfig
-    *   the configuration file
+    *   The configuration for SIMONA
     * @return
-    *   the resulting result file hierarchy
+    *   The resulting result file hierarchy
```
sebastian-peter, line 234: Our new formatting rules also include full stops at the end of text lines, so it'd look something like this: 

```suggestion
    * [[SimonaConfig]], otherwise an exception is thrown.
    *
    * @param typeSafeConfig
    *   All configuration parameters.
    * @param simonaConfig
    *   The configuration for SIMONA.
    * @return
    *   The resulting result file hierarchy.
```

Just something I wanted to mention, because then the format does not have to be touched again :)

sebastian-peter (changes requested):
One tiny point, otherwise LGTM

PR #1304, opened 2025-03-24
Made some method of `PvModel` static

### LINKED ISSUE #1217: Move static methods in `PvModel` class to the companion object
Since these methods should not depend on the fields of the class, but rather get all the relevant data via the method parameters, they can also be static (thus in the object) and therefore be safer against future mishaps.

staudtMarius:
Resolves #1217


sebastian-peter (approved):
LGTM, just added a return type

PR #1316, opened 2025-03-27
Add conversion utils between `ComparableQuantity` and `Squants`.

### LINKED ISSUE #1313: Add conversion utils between psdm and simona quantities

staudtMarius:
Resolves #1313


src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
sebastian-peter, line 1: Maybe we can generalize this a bit more - we're not only converting e.g. kilo volt amperes from one to the other library, but apparent power. Furthermore, the way this is done right now, it could be confusing because e.g. `toKilovoltamperes` and similar already exists in the context of squants.

So, a more fitting name would maybe be something like `toSquants` for all methods, and then you would use the base unit (volt amperes here) for conversion. Let me know what you think.

sebastian-peter (changes requested):
Really like the initiative, this looks a lot cleaner. Please have a look at my discussion point.

src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
staudtMarius: I change the name a bit. I would still keep using the unit, we defined in the PSDM (e.g. kVA).

sebastian-peter: Seems fine to me, unless there's other opinions @danielfeismann @t-ober

src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
```diff
@@ -0,0 +1,231 @@
...
+      */
+    def toPu: squants.Dimensionless = Each(
+      quantity.to(PU).getValue.doubleValue
+    )
+
+    /** Returns a quantity with unit [[Percent]].
+      */
+    def toPercent: squants.Dimensionless = Percent(
+      quantity.to(PERCENT).getValue.doubleValue
+    )
```
sebastian-peter, line 48: Reminder that this can be replaced by toSquants

sebastian-peter (approved):
LGTM, thanks!

PR #1323, opened 2025-04-01
Fix determining FlexOptions of Evcs

### LINKED ISSUE #1322: Fix determining FlexOptions of Evcs
We determine preferred(Charging)Power for all evs including the those that are already full

https://github.com/ie3-institute/simona/blob/16aaac389ee8d3a28e463962d13cd5b5978533ab/src/main/scala/edu/ie3/simona/model/participant2/evcs/EvcsModel.scala#L228-L233

this should only be done for the not fully charged evs, since we filter only for them when determine the operating point

https://github.com/ie3-institute/simona/blob/16aaac389ee8d3a28e463962d13cd5b5978533ab/src/main/scala/edu/ie3/simona/model/participant2/evcs/EvcsModel.scala#L302-L308

An alternative solution would be to include the toleranceMargin for all charging strategies

https://github.com/ie3-institute/simona/blob/16aaac389ee8d3a28e463962d13cd5b5978533ab/src/main/scala/edu/ie3/simona/model/participant2/evcs/MaximumPowerCharging.scala#L19-L24

danielfeismann:
resolve #1322


sebastian-peter:
@danielfeismann Added the test, have a look if you like

danielfeismann (commented):
Thanks for addding the test. LGTM as well

sebastian-peter (approved):

PR #1325, opened 2025-04-01
Introducing participant model factory pattern

### LINKED ISSUE #1324: Introduce ParticipantModel factory pattern
This should allow for proper integration of the LoadProfileService with #1235

sebastian-peter:
Resolves #1324

Merge first:
-  #1328


src/main/scala/edu/ie3/simona/model/participant2/load/FixedLoadModel.scala
```diff
@@ -41,36 +46,53 @@ class FixedLoadModel(
...
+
+    override def getInitialState(
+        tick: Long,
+        simulationTime: ZonedDateTime,
+    ): FixedState = FixedState(tick)
+
+    override def create(): ParticipantModel[
+      _ <: OperatingPoint,
+      FixedState,
+    ] = {
```
staudtMarius, line 66: I think this should be replaced with `FixedLoadModel`.

src/test/scala/edu/ie3/simona/agent/participant2/ParticipantAgentInitSpec.scala
```diff
@@ -177,11 +177,11 @@ class ParticipantAgentInitSpec
 
         participantAgent ! PrimaryRegistrationSuccessfulMessage(
           primaryService.ref,
-          15 * 3600L,
+          4 * 3600L,
```
staudtMarius, line 180: Why did you change this? Also at the other places in this file.

staudtMarius (changes requested):
LGTM. I have just two small questions.

src/test/scala/edu/ie3/simona/agent/participant2/ParticipantAgentInitSpec.scala
sebastian-peter: After discussion with @danielfeismann, we decided that if data (primary and secondary) has not been received before the start of the operation interval, that's an error, and we're throwing. Beforehand, we would just postpone the start of the operation, which is something you probably wouldn't expect.

staudtMarius (approved):
LGTM.

PR #1329, opened 2025-04-04
New hp model with operating points for thermalGrid

danielfeismann:


src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -586,16 +587,16 @@ class ThermalGridIT
       // point nor one of it reached any boundary.
       resultListener.expectNoMessage()
 
-      scheduler.expectMessage(Completion(heatPumpAgent, Some(31940)))
+      scheduler.expectMessage(Completion(heatPumpAgent, Some(31939)))
```
sebastian-peter, line 590: Just to be sure: What's the cause of this?

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -735,7 +736,7 @@ class ThermalGridIT
             }
         }
 
-      scheduler.expectMessage(Completion(heatPumpAgent, Some(71891)))
+      scheduler.expectMessage(Completion(heatPumpAgent, Some(71892)))
```
sebastian-peter, line 739: And here, tick is even increased instead of decreased...

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -996,7 +997,9 @@ class ThermalGridIT
                 emResult._3 should equalWithTolerance(
                   -0.002517561515.asMegaWatt
                 )
-                emResult._4 should equalWithTolerance(0.0012489996.asMegaVar)
+                emResult._4 should equalWithTolerance(
+                  -0.00082748245392177.asMegaVar
+                )
```
sebastian-peter, line 1002: Also this looks like an entirely different value...

sebastian-peter (changes requested):
Looks nice and _definitely_ cleaner than before!

Haven't gone through all tests in detail, can do so though if you think that'd be a good idea.

Edit: Maybe it'd be an idea to integrate the thermal losses of the hosue into the operating point? That way, we could provide a _true_ "zero" operating point

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -288,27 +280,22 @@ final case class ThermalGrid(
...
       state: HpState,
       qDotHouse: Power,
-  ): (Option[ThermalHouseState], Option[ThermalThreshold], Power) = {
+  ): (Power, Option[ThermalThreshold]) = {
     (house, state.thermalGridState.houseState) match {
       case (Some(thermalHouse), Some(lastHouseState)) =>
-        val (newState, threshold) = thermalHouse.updateState(
-          tick,
+        val newState = thermalHouse.determineState(
+          state.tick,
```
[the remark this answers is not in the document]
danielfeismann: Right. Good catch!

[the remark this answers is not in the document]
danielfeismann: And even identified a failure in the tests

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
danielfeismann: That one is caused by line 74-75 where I changed the tolerance. Which was a dangerous one we missed in the previous PRs. I can fix that in a separate PR if you like for better traceability?

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
danielfeismann: Same here, for better traceability we could move this in a separate PR upfront, if you like.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
danielfeismann: Right. I followed this backwards:

1. one tick difference in variable duration in `ThermalHouse.nextActivation` caused by 
2. slightly different qDot (8e-6) and slightly different `flexibleEnergy` (what a name) (6e-4), the latter caused by
3. different innerTemp because, my assumption, we calculate at ticks that are a little bit different and round there also a little bit. 

imho this is ok, but yes, over time this error might add up a little bit. Which it always has been, right? Or am I missing something?

danielfeismann:
Thanks for your review!

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
sebastian-peter: Well the difference between `1.2489996 kVar` and `-0.82748245392177 kVar` surely seems to be _not_ caused by a changed testing tolerance, right?

sebastian-peter: Ah, ok, I see. It was the `0.01` implicit double.

danielfeismann:
Regarding the tests, maybe you could also have a look if I missed some test case / have something twice in there, some test case no longer make any sense... will also check that again.

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
```diff
@@ -526,10 +764,102 @@ class HpModelSpec
...
+      val ambientTemperature = Celsius(10)
+
+      val cases = Table(
+        (
+          "tick",
+          "requiredDemandHouse",
+          "expectedHpQDot",
+          "expectedTick",
+        ),
+        //  (0, 0d, 0d, Some(4000)),
```
sebastian-peter, line 780: Some more cases here?

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
```diff
@@ -526,10 +764,102 @@ class HpModelSpec
               minPower should approximate(Kilowatts(expectedMinPower))
               maxPower should approximate(Kilowatts(expectedMaxPower))
           }
+      }
+    }
+
+    "Handle operating correctly" in {
```
sebastian-peter, line 770:
```suggestion
    "determine operating point without flex control correctly" in {
```

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
```diff
@@ -526,10 +764,102 @@ class HpModelSpec
...
+          val (op, threshold) = hpModel.determineOperatingPoint(state)
+
+          op.activePower shouldBe Kilowatts(expectedHpQDot)
+          threshold shouldBe expectedTick
 
       }
     }
 
-    "Handle controlled power change" in {}
+    "Handle controlled power change" in {
```
sebastian-peter, line 815:
```suggestion
    "determine operating point with flex control correctly" in {
```

sebastian-peter (changes requested):
Some more points, still looking through the other tests...

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
sebastian-peter: Regarding capital first letter: Just realized you use a different naming structure. It was supposed to work like this:

`HpModel` should `determine whatever` in {}

But Intellij doesn't display it like this, so to me both is fine

src/test/scala/edu/ie3/simona/model/participant2/HpModelSpec.scala
danielfeismann: Oh, I missed this one! 

But I see just the two cases, do you have any other case in mind?

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithStorageOnlySpec.scala
```diff
@@ -130,18 +132,16 @@ class ThermalGridWithStorageOnlySpec
...
-          testGridAmbientTemperature,
+          HpOperatingPoint(zeroKW, ThermalGridOperatingPoint.zero),
           onlyAdditionalDemandOfHeatStorage,
         )
 
         val updatedThermalGridState =
-          thermalGrid.updatedThermalGridState(
+          thermalGrid.determineThermalGridState(
             state.tick,
             state,
```
sebastian-peter, line 143: This does nothing, right? Last state is given at the current tick

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
```diff
@@ -276,381 +350,224 @@ class ThermalGridWithHouseAndStorageSpec
...
+              HouseTemperatureLowerBoundaryReached(154285)
             )
-            nextThreshold shouldBe None
         }
       }
 
-      "hand back unaltered information if house temperature is above lower boundary temperature" in {
-        val maybeHouseState = Some(
+      "heat house from storage if house temperature is at lower boundary temperature" in {
+        val maybeHouseState =
```
sebastian-peter, line 415: Value naming

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
```diff
@@ -276,381 +350,224 @@ class ThermalGridWithHouseAndStorageSpec
...
-        ) match {
-          case (revisedGridState, nextThreshold) =>
-            revisedGridState shouldBe ThermalGridState(
-              maybeHouseState,
-              maybeStorageState,
-            )
-            nextThreshold shouldBe None
-        }
-      }
+        val maybeStorageState =
```
sebastian-peter, line 430: Could also be a storageState and apply Some() later

sebastian-peter (changes requested):
Some more or less nitty last requests from my side, with the more substantial ones being these:
- `state: HpState` should be `lastState: HpState` in determineState
- In ThermalGrid, `determineThermalGridState` also takes a `lastState`. ScalaDoc could be adapted accordingly. Also, what should "actual" mean? Probably should say "current" instead.
- Similar issues in `ThermalHouse.determineState`

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithStorageOnlySpec.scala
danielfeismann: Right 😄

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -48,27 +48,27 @@ final case class ThermalGrid(
     heatStorage: Option[ThermalStorage],
 ) extends LazyLogging {
 
-  /** Determines the state of the ThermalGrid by using the actual
+  /** Determines the state of the ThermalGrid by using the current
```
sebastian-peter, line 51: Actually, "current" might be misleading here, since it's an operating point determined in the past...

sebastian-peter (changes requested):
Ok, only this point besides the other below missing:
- `state: HpState` should be `lastState: HpState` in HpModel.determineState (sorry, didn't communicate where)

sebastian-peter (approved):

PR #1337, opened 2025-04-08
Refactoring uuids in default runtime configs.

### LINKED ISSUE #1336: Refactor uuids in default runtime configs.

staudtMarius:
Resolves #1336


src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
```diff
@@ -596,86 +596,6 @@ class ConfigFailFastSpec extends UnitSpec with ConfigTestData {
...
-          intercept[InvalidConfigParameterException] {
-            ConfigFailFast invokePrivate checkBaseRuntimeConfigs(
-              simonaConfig.simona.runtime.participant.load.defaultConfig,
-              simonaConfig.simona.runtime.participant.load.individualConfigs,
-              defaultString,
-            )
-          }.getMessage shouldBe "Found invalid UUID 'blabla' it was meant to be the string 'default' or a valid UUID."
-        }
-
-        "throw an InvalidConfigParameterException if the UUID of the base model config is not valid" in {
```
sebastian-peter, line 653: This test would be still helpful, wouldn't it?

sebastian-peter (changes requested):

src/test/scala/edu/ie3/simona/config/ConfigFailFastSpec.scala
staudtMarius: Since we no longer need a `default` string, the elements are no longer checked in `ConfigFailFast`. With this PR only a log statement is made, if the list of uuids is not empty.

sebastian-peter: The way I see it this test was about malformed UUIDs though, right?

staudtMarius: Yes, but the test only checks the uuid of the base (default) config. I will adapt the test, to test the uuids of the individual configs.

sebastian-peter (changes requested):
One more test that could be preserved, in my opinion...

sebastian-peter (approved):
Thanks!

PR #1343, opened 2025-04-11
Refactor handleFeedIn and handleConsumption of thermalGrid due to qDo…

### LINKED ISSUE #1342: Refactor handleFeedIn and handleConsumption of thermalGrid due to qDotIntoGrid
We're a bit inconsistent here:

https://github.com/ie3-institute/simona/blob/29ae1e7ff2f3efbf36f9cc206ee92b650db651b5/src/main/scala/edu/ie3/simona/model/participant2/HpModel.scala#L181-L196

when the heatPump is not running, the qDot provided by `HeatPump` would be zero. What is really ment is the some qDot into the grid. But secondly, this is already "in the grid" since it's qDot provided by the thermal storage. Thus, this should be handled via `ThermalGrid.handleConsumption` not `ThermalGrid.handleFeedin` later down the road...

danielfeismann:
…tIntoGrid

resolves #1342

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Isolate solar radiation calculations of PvModel to its own object [#1327](https://github.com/ie3-institute/simona/issues/1327)
 - Reorganized CI order [#1333](https://github.com/ie3-institute/simona/issues/1333)
 - Fixed determining FlexOptions of `Evcs` [#1322](https://github.com/ie3-institute/simona/issues/1322)
+- Refactor handleFeedIn and handleConsumption of `ThermalGrid` due to qDotIntoGrid [#1343](https://github.com/ie3-institute/simona/issues/1343)
 
 ### Fixed
 - Fix rendering of references in documentation [#505](https://github.com/ie3-institute/simona/issues/505)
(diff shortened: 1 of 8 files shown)


danielfeismann:
> Looks fine to me, although I'm not 100% clear on why this works without the complex code in `handleFeedIn`.
> 
> I have some minor other points, not directly related to the PR, but maybe you could handle them since you're on it:
> 
>     * In `ThermalGrid`, `state` is often times described as "Last state of the heat pump", which is not correct.
> 
>     * `handleCases` should be called `handleCase` imho, since the case to be handled has already been decided beforehand - we're only dealing with a single case at a time.
> 
>     * `handleStorageCases` is doing something that would, after [New hp model with operating points for thermalGrid #1329](https://github.com/ie3-institute/simona/pull/1329), be probably called something like `determineStorageOperatingPoint`? I guess this could be handled then, not here.

At first I also wondered how it could be so simple. But it makes sense for two reasons. Firstly, moving the feed in from storage within `HpModel` to `revisedFeedInFromStorage` makes it unnecessary to check for these cases in `handleFeedIn` as this belongs to `handleConsumption` and was partly already handled there.  Second, the parts where we want to behave like lastState are identical to what the behaviour would be later on. So this is also redundant.

Good suggestions, I will include them. I guess, you're right with storageOperatingPoint but that should be done then.

sebastian-peter (approved):

PR #1346, opened 2025-04-15
Fix tolerance in ThermalGridIT

### LINKED ISSUE #1345: Fix tolerance in ThermalGridIT
https://github.com/ie3-institute/simona/blob/3cd4435fe6a429f65a4afe54a94d054a0019764d/src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala#L74

this should not be implict.
There should be an implict tolerance of e.g. e-10

danielfeismann:
resolves #1345


sebastian-peter (approved):
Ok, so we were not actually checking results all along... Thanks for fixing this!

PR #1356, opened 2025-04-22
Fixing default config values not written to output config file.

### LINKED ISSUE #1355: Write out default config parameters
Currently, only the config parameters set in the `.conf` file are written. With this issue all default config parameter, that are not overwritten, should also be added to the output config.

staudtMarius:
Resolves #1355


danielfeismann (approved):
LGTM

PR #1350, opened 2025-04-17
Calculate ThermalHouse new inner temperature correct 

danielfeismann:
resolves #1349 

Credits also to @jo-bao 

merge first
- #1329

PR #1358, opened 2025-04-22
Introducing energy demand for warm water heating

### LINKED ISSUE #856: Enhance thermal house model by heat demand for tap water

danielfeismann:
resolves #856 

merge first
- #1350
- #1392
- #1474 
- #1476
- #1478
- #1482
- #1548 
- #1555
- #1557 
- #1564


src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -15,11 +15,11 @@ import edu.ie3.datamodel.models.input.thermal.{
 import edu.ie3.simona.model.participant.ParticipantModel.ModelState
 import edu.ie3.simona.model.participant.hp.HpModel.HpState
 import edu.ie3.simona.model.thermal.ThermalGrid.ThermalEnergyDemand
+import edu.ie3.simona.model.thermal.ThermalHouse.*
 import edu.ie3.simona.model.thermal.ThermalHouse.ThermalHouseThreshold.{
   HouseTargetTemperatureReached,
   HouseTemperatureLowerBoundaryReached,
 }
```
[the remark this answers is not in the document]
danielfeismann: Would be great if spotless would cover that

docs/readthedocs/models/thermal_house_model.md
```diff
@@ -7,8 +7,7 @@ This page documents the functionality of the thermal house available in SIMONA.
 
 ## Behaviour
 
-This house model represents the thermal behaviour of a building. It represents a simple shoebox with thermal capacity and transmission losses.
-The house can optionally be equipped with a {ref}`cts_model` as thermal storage. Both are connected by the {ref}`thermal_grid_model`.
+This house model represents the thermal behaviour of a building. It represents a simple shoebox with thermal capacity and  transmission losses for its heating demand. As well the hot water demand (tap water) based on the number of inhabitants and the housing type on hourly basis is considered. The house can optionally be equipped with a {ref}`cts_model` as thermal storage. Both are connected by the {ref}`thermal_grid_model`.
```
staudtMarius: You added an additional space between `and` and `transmission` that can be removed.

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
```diff
@@ -78,7 +81,21 @@ class HpModel private (
...
-    val thermalDemands = thermalGrid.determineEnergyDemand(thermalGridState)
+    val hoursWaterDemandToDetermine = thermalGrid.house match {
+      case Some(house) =>
+        house.checkIfNeedToDetermineDomesticHotWaterDemand(
+          tick,
+          simulationTime,
+          lastState,
+        )
+      case None => None
+    }
```
staudtMarius: Could be done in a `map` statement, since the `None` case returns `None`. 
It could look like: `.map(_.checkIfNeedToDetermineDomesticHotWaterDemand(...))`

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
```diff
@@ -246,19 +263,23 @@ class HpModel private (
...
+      thermalDemands.domesticHotWaterStorageDemand
     val noHeatStorageOrEmpty = thermalGridState.isHeatStorageEmpty
 
     val turnHpOn =
       (demandHouse.hasRequiredDemand && noHeatStorageOrEmpty) ||
         (demandHouse.hasPossibleDemand && wasRunningLastPeriod ||
           demandHeatStorage.hasRequiredDemand ||
-          (demandHeatStorage.hasPossibleDemand && wasRunningLastPeriod))
+          (demandHeatStorage.hasPossibleDemand && wasRunningLastPeriod)) ||
+        demandDomesticHotWaterStorage.hasRequiredDemand
```
staudtMarius, line 273: DId you forget the check for `demandDomesticHotWaterStorage.hasPossibleDemand && wasRunningLastPeriod`?
Also, why is this check no inside the previous bracket, while the normal heat storage is?

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
```diff
@@ -246,19 +263,23 @@ class HpModel private (
...
 
     val canOperate =
       demandHouse.hasRequiredDemand || demandHouse.hasPossibleDemand ||
-        demandHeatStorage.hasRequiredDemand || demandHeatStorage.hasPossibleDemand
+        demandHeatStorage.hasRequiredDemand || demandHeatStorage.hasPossibleDemand ||
+        demandDomesticHotWaterStorage.hasRequiredDemand
```
staudtMarius, line 278: Why are you only checking for required demand here, but no the possible demand?

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
```diff
@@ -333,16 +354,21 @@ object HpModel {
     *   [[edu.ie3.simona.model.thermal.ThermalHouse]] used for space heating.
     * @param qDotHeatStorage
     *   The thermal power input of the
-    *   [[edu.ie3.simona.model.thermal.ThermalStorage]].
+    *   [[edu.ie3.simona.model.thermal.ThermalStorage]] used for heat storage.
+    * @param qDotDomesticHotWaterStorage
+    *   The thermal power input of the
+    *   [[edu.ie3.simona.model.thermal.ThermalHouse]] used for domestic hot
```
staudtMarius: Is this the correct class? Shouldn't this be the class of the domestic hot water storage?

src/main/scala/edu/ie3/simona/model/thermal/ThermalDemandConditions.scala
```diff
@@ -38,14 +40,19 @@ private object ThermalDemandConditions {
     ThermalDemandConditions(
       /* Consider the action in the last state
        * We can continue using the qDots from last operating point to keep continuity.
-       * If the house was heated in lastState and has still some demand.
-       */
+       * If the house was heated in lastState and has still some demand and the domestic
+       * hot water storage has no demand. */
```
staudtMarius, line 44: Why are we stopping the heating if the hot water storage has some demand?

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
```diff
@@ -510,6 +510,90 @@ class EmAgentIT
...
+            emResult.getP should equalWithTolerance(-0.0055734002706.asMegaWatt)
+            emResult.getQ should equalWithTolerance(-0.00183188808074.asMegaVar)
+        }
+        resultListener.expectNoMessage()
+        scheduler.expectMessage(Completion(emAgentActivation, Some(3600)))
+
+        /* TICK 3600
+        DomesticHotWaterStorage stopped discharging. Expect same behaviour as before
+        LOAD: 0.269 kW (unchanged)
+        PV:  -5.842 kW FIXME
```
staudtMarius: Will this be fixed in another issue?

staudtMarius (changes requested):
Here are some comments from my side. I will have a closer look at the tests later.

src/main/scala/edu/ie3/simona/model/thermal/ThermalDemandConditions.scala
danielfeismann: We don't stop. This only checks if we can continue heating the house as in last time step. Therefore we check:
1. was house heated last step
2. does it still has some demand
3. has the hot water storage NO demand (if it would have some, we need to change the heat distribution and share between house and water storage)

src/test/scala/edu/ie3/simona/agent/em/EmAgentIT.scala
danielfeismann: No, this was only a reminder to check the pv power in the end which I missed... Thanks, will fix this.

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
danielfeismann: The idea was and is to keep the domestic hot water storage out of the flexibility patterns. This can be discussed but I would like to keep it simple as it is still complicated enough... I added some phrase to the docs.

src/main/scala/edu/ie3/simona/model/participant/hp/HpModel.scala
danielfeismann: What do you mean be the check for `demandDomesticHotWaterStorage.hasPossibleDemand && wasRunningLastPeriod`?

But one of the brackets was wrong. Good catch!

staudtMarius (approved):

PR #1363, opened 2025-04-28
Fix some default values not being used.

### LINKED ISSUE #1362: Fix some default config values not being used.
There is an issue with the `.empty` methods in the config. Because the default values are not directly set in the definition of the case class, currently, you have to configure all values.

staudtMarius:
Resolves #1362


danielfeismann (approved):
LGTM

PR #1366, opened 2025-05-07
Making primary data source optional.

### LINKED ISSUE #134: Config - Make primary data source optional
As of now the config expects a primary data source even when you do not want to use any. Make the config parameter optional and adapt accordingly.

staudtMarius:
Resolves #134


sebastian-peter (approved):
LGTM

PR #1392, opened 2025-05-20
Adapt ThermalHouse and HP flexibility behaviour

### LINKED ISSUE #1391: Adapt ThermalHouse and HP flexibility behaviour
In case of an EM controlled thermal house and heat pump (strategy selfConsumption), i would like to change some specific behaviour.

If the house is heated by the heat pump and then, for example, the weather changes so that the EM setPower is zero or below the value required to operate the heat pump, the heat pump would stop heating the house. While this is correct from the perspective of a strict flexibility strategy, it also means that the house would only be heated to the target temperature when there is surplus energy (setPower > 0.5 sRated of HP for the selfConsumption strategy) or when the thermal storage has energy stored in it. This only happens when there was surplus energy before.

I would like to add the following condition: the flexOptions of the HP should change to referencePower == runningHp if all of the following conditions are true:
- the HP was running in the last state
- the house has a demand
- the thermal storage is empty (or does not exist).

This would change it in a way, that the house is not interrupted in some heating process due to weather change. For me this is a compromise between always heat back to target temperature when reaching the lower temperature (resulting in only positive flex that can be provided by the Hp) and the actual behaviour that could lead to cases where the house stays at the lower temperature most of the time.

I'm open for discussion on this. @sebastian-peter @staudtMarius what do you think?

danielfeismann:
resolves #1391


src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
```diff
@@ -114,11 +114,39 @@
...
+      case (true, false) =>
+        (sRated.toActivePower(cosPhiRated), sRated.toActivePower(cosPhiRated))
+      case (false, true) =>
+        (
+          zeroKW,
+          zeroKW,
+        )
+      case (false, false) =>
+        (
+          sRated.toActivePower(cosPhiRated),
```
staudtMarius: I think, this should be `zeroKW`. Since `turnOn` is false, the ref power should be zero.

staudtMarius (changes requested):
Looks good to me. I just have one small question regarding new the determination of the ref power.

src/main/scala/edu/ie3/simona/model/participant/HpModel.scala
danielfeismann: You're right in some sense, but this case is anyway not possible. But an exception might be the better solution.

staudtMarius (dismissed):
LGTM. @sebastian-peter if you don't have any objections, this PR can be merged.

src/main/scala/edu/ie3/simona/model/participant/hp/HpPowerLimitFlexModel.scala
```diff
@@ -30,11 +29,46 @@ class HpPowerLimitFlexModel(private val model: HpModel)
         wasRunningLastOp,
       )
 
-    PowerLimitFlexOptions(
-      if (turnOn) model.sRated.toActivePower(model.cosPhiRated) else zeroKW,
-      if (canBeOutOfOperation) zeroKW else model.pRated,
-      if (canOperate) model.pRated else zeroKW,
-    )
+    val (refPower, minPower) = (turnOn, canBeOutOfOperation) match {
```
sebastian-peter, line 41: Just a remark on code style here. Maybe one could declare a boolean tuple `val (refOn, minOn)` here, whose values get turned into `model.sRated.toActivePower(model.cosPhiRated)` or `zeroKW` later. This way, one could reduce redundant code a bit.

sebastian-peter (commented):
Small comment from my side

src/main/scala/edu/ie3/simona/model/participant/hp/HpPowerLimitFlexModel.scala
danielfeismann: not sure if i got you complete but did some refactoring. Please have a look

sebastian-peter: This is not what I meant exactly, but it works as well. Thanks

src/main/scala/edu/ie3/simona/model/participant/hp/HpPowerLimitFlexModel.scala
```diff
@@ -28,47 +28,41 @@ class HpPowerLimitFlexModel(private val model: HpModel)
...
-        throw IllegalStateException(
+        throw new IllegalStateException(
           "An unsupported FlexOption for a heat pump has been determined."
         )
       // should not be possible to reach
     }
 
     val maxPower =
-      if (canOperate) model.sRated.toActivePower(model.cosPhiRated) else zeroKW
+      if canOperate then model.sRated.toActivePower(model.cosPhiRated)
```
sebastian-peter: One more line where you could replace with `refOn` :)

sebastian-peter (approved):

sebastian-peter (approved):

PR #1399, opened 2025-05-21
Fixes in Documentation, ScalaDocs, Code Style 

### LINKED ISSUE #1397: Documentation, ScalaDoc and Code Style fixes
Mostly format and other stuff, but from content side there will be changes of
- pattern matching conditions for writting results of ThermalGrid
- additional checks that resultListener expect no more messages in ThermalGridIT and EmAgentIT

danielfeismann:
resolves #1397

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
@@ -165,20 +165,19 @@ object CylindricalThermalStorage {
       input: CylindricalStorageInput,
       initialStoredEnergy: Energy = zeroKWh,
   ): CylindricalThermalStorage = {
-    val maxEnergyThreshold =
-      CylindricalThermalStorage.volumeToEnergy(
-        CubicMeters(
-          input.getStorageVolumeLvl.to(Units.CUBIC_METRE).getValue.doubleValue
-        ),
-        KilowattHoursPerKelvinCubicMeters(
-          input.getC
-            .to(PowerSystemUnits.KILOWATTHOUR_PER_KELVIN_TIMES_CUBICMETRE)
-            .getValue
-            .doubleValue
-        ),
-        Celsius(input.getInletTemp.to(Units.CELSIUS).getValue.doubleValue()),
-        Celsius(input.getReturnTemp.to(Units.CELSIUS).getValue.doubleValue()),
-      )
+    val maxEnergyThreshold = volumeToEnergy(
+      CubicMeters(
+        input.getStorageVolumeLvl.to(Units.CUBIC_METRE).getValue.doubleValue
+      ),
+      KilowattHoursPerKelvinCubicMeters(
+        input.getC
+          .to(PowerSystemUnits.KILOWATTHOUR_PER_KELVIN_TIMES_CUBICMETRE)
+          .getValue
+          .doubleValue
+      ),
+      Celsius(input.getInletTemp.to(Units.CELSIUS).getValue.doubleValue),
+      Celsius(input.getReturnTemp.to(Units.CELSIUS).getValue.doubleValue),
+    )
 
     val pThermalMax = Kilowatts(
       input
--- a/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
@@ -175,8 +175,8 @@ object CylindricalThermalStorage {
           .getValue
           .doubleValue
       ),
-      input.getInletTemp.toSquantsCelsius,
-      input.getReturnTemp.toSquantsCelsius,
+      input.getInletTemp.toSquants,
+      input.getReturnTemp.toSquants,
     )
 
     val pThermalMax = Kilowatts(
--- a/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
@@ -19,7 +19,11 @@ import edu.ie3.simona.model.thermal.ThermalStorage.ThermalStorageThreshold.{
 }
 import edu.ie3.util.quantities.PowerSystemUnits
 import edu.ie3.util.scala.quantities.DefaultQuantities.*
-import edu.ie3.util.scala.quantities.QuantityConversionUtils.TemperatureConversionSimona
+import edu.ie3.util.scala.quantities.QuantityConversionUtils.{
+  TemperatureConversionSimona,
+  VolumeConversionSimona,
+  PowerConversionSimona,
+}
 import edu.ie3.util.scala.quantities.SquantsUtils.RichEnergy
 import edu.ie3.util.scala.quantities.{
   KilowattHoursPerKelvinCubicMeters,
@@ -166,9 +170,7 @@ object CylindricalThermalStorage {
       initialStoredEnergy: Energy = zeroKWh,
   ): CylindricalThermalStorage = {
     val maxEnergyThreshold = volumeToEnergy(
-      CubicMeters(
-        input.getStorageVolumeLvl.to(Units.CUBIC_METRE).getValue.doubleValue
-      ),
+      input.getStorageVolumeLvl.toSquants,
       KilowattHoursPerKelvinCubicMeters(
         input.getC
           .to(PowerSystemUnits.KILOWATTHOUR_PER_KELVIN_TIMES_CUBICMETRE)
@@ -179,13 +181,7 @@ object CylindricalThermalStorage {
       input.getReturnTemp.toSquants,
     )
 
-    val pThermalMax = Kilowatts(
-      input
-        .getpThermalMax()
-        .to(PowerSystemUnits.KILOWATT)
-        .getValue
-        .doubleValue()
-    )
+    val pThermalMax = input.getpThermalMax().toSquants
 
     new CylindricalThermalStorage(
       input.getUuid,
--- a/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
@@ -27,7 +27,6 @@ import edu.ie3.util.scala.quantities.{
 }
 import squants.energy.Kilowatts
 import squants.space.{CubicMeters, Volume}
-import squants.thermal.Celsius
 import squants.time.Seconds
 import squants.{Energy, Power, Temperature}
 import tech.units.indriya.unit.Units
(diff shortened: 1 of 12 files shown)


src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
```diff
@@ -165,21 +165,19 @@ object CylindricalThermalStorage {
...
+      ),
+      KilowattHoursPerKelvinCubicMeters(
+        input.getC
+          .to(PowerSystemUnits.KILOWATTHOUR_PER_KELVIN_TIMES_CUBICMETRE)
+          .getValue
+          .doubleValue
+      ),
+      Celsius(input.getInletTemp.to(Units.CELSIUS).getValue.doubleValue),
+      Celsius(input.getReturnTemp.to(Units.CELSIUS).getValue.doubleValue),
+    )
```
staudtMarius: Maybe this could be moved to the quantity to squants conversion utils.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -321,15 +318,14 @@ final case class ThermalGrid(
   }
 
   /** Handles the case, when the storage has heat demand and will be filled up
-    * here (positive qDot) or will return its stored energy into the thermal
-    * grid (negative qDot).
+    * here (positive qDot).
```
staudtMarius, line 321: Is a negative qDot still possible here? Since below you state, that `qDotStorage` can be negative.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -517,12 +511,12 @@ final case class ThermalGrid(
...
         lastOpThermals.forall(_.qDotHouse != currentOpThermals.qDotHouse),
-        state.tick != 0,
+        state.tick == 0,
       ) match {
-        case (Some(house: ThermalHouse), true, true) =>
+        case (Some(house: ThermalHouse), true, _) =>
           createThermalHouseResult(house)
         // We always want the results of the first tick
-        case (Some(house: ThermalHouse), _, false) =>
+        case (Some(house: ThermalHouse), _, true) =>
```
staudtMarius: Why do we need these two cases? They will both return the same result.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -534,12 +528,12 @@ final case class ThermalGrid(
...
         ),
-        state.tick != 0,
+        state.tick == 0,
       ) match {
-        case (Some(storage: CylindricalThermalStorage), true, true) =>
+        case (Some(storage: CylindricalThermalStorage), true, _) =>
           createCylindricalStorageResult(storage)
         // We always want the results of the first tick
-        case (Some(storage: CylindricalThermalStorage), _, false) =>
+        case (Some(storage: CylindricalThermalStorage), _, true) =>
```
staudtMarius: Same as above.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -493,24 +480,22 @@ class ThermalGridIT
...
       // Since this activation is caused by new weather data, we don't expect any
       // message for house or storage since there is no change of their operating
       // point nor one of it reached any boundary.
       resultListener.expectNoMessage()
-
       scheduler.expectMessage(Completion(heatPumpAgent, Some(26887)))
 
       /* TICK 26887
-        Storage will be empty
+        Heat storage is empty
```
staudtMarius, line 494: Before, you changed is to `ThermalStorage`. Which one is correct?

staudtMarius (changes requested):
Looks already quiete good. Just some small remarks from my side.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
danielfeismann: This is a heat storage, there is also the domestic hot water storage, both are ThermalStorages. So heat storage should be correct.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
danielfeismann: We only would like to write a thermalResult either when 
- there is a change in the result (will be, if the qDot changes)
- in any case if we are at tick 0

Do you see a better option to achieve this?

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
danielfeismann: It could be possible here (and would work fine since `determineNextThreshold` handles both cases correct), but since this is only called by `handleCases` which only is called for feedIn, there won't be any negative qDot here.

I kept the qDotStorage scalaDoc unchanged since it is still valid and helps for understanding.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
staudtMarius: Maybe we can write the condition as `lastOpThermals.forall(_.qDotHouse != currentOpThermals.qDotHouse) || state.tick == 0`.
With this, there would only be one case.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
staudtMarius: Ok. But then you should rename the `storage` in line 242, 990, 1239 and 1859 into `heat storage` as well.

src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
```diff
@@ -165,21 +165,19 @@ object CylindricalThermalStorage {
...
+    val maxEnergyThreshold = volumeToEnergy(
+      CubicMeters(
+        input.getStorageVolumeLvl.to(Units.CUBIC_METRE).getValue.doubleValue
+      ),
+      KilowattHoursPerKelvinCubicMeters(
+        input.getC
+          .to(PowerSystemUnits.KILOWATTHOUR_PER_KELVIN_TIMES_CUBICMETRE)
+          .getValue
+          .doubleValue
+      ),
```
staudtMarius: In my opinion, it would be better to also move these two conversion to the utils class.

src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
```diff
@@ -243,4 +242,27 @@ object QuantityConversionUtils {
...
+      */
+    def toSquantsCelsius: squants.Temperature = Celsius(
+      quantity.to(CELSIUS).getValue.doubleValue
+    )
+
+    /** Returns a quantity with unit [[Kelvin]].
+      */
+    def toSquantsKelvin: squants.Temperature = Kelvin(
+      quantity.to(KELVIN).getValue.doubleValue
+    )
```
staudtMarius: Do we need the second method? If not, it can be removed and the first method could be renamed to `toSquants` like most of the conversion methods.

src/test/scala/edu/ie3/simona/agent/grid/ThermalGridIT.scala
```diff
@@ -236,16 +235,15 @@ class ThermalGridIT
                 energy should equalWithTolerance(0.asMegaWattHour)
             }
         }
-
+      resultListener.expectNoMessage()
       scheduler.expectMessage(Completion(heatPumpAgent, Some(3416)))
 
       /* TICK 3416
-      Storage is fully heated up
+      ThermalStorage is fully heated up
```
staudtMarius: This could be renamed into `heat storage`. There are a few more such cases.

staudtMarius (changes requested):
Some small questions.

src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
danielfeismann: To be honest, I won't do this since I guess we have this conversion only once at this place.

src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
danielfeismann: I remember that it was sometimes, no idea why, important to convert to Celsius or Kelvin. Therefor I guess we need both.

src/main/scala/edu/ie3/simona/model/thermal/CylindricalThermalStorage.scala
staudtMarius: Ok. My idea was, that it would look cleaner. But we can leave it as it is now.

src/main/scala/edu/ie3/util/scala/quantities/QuantityConversionUtils.scala
staudtMarius: But we are using squants. We can convert the quantity to Kelvin, if we need it at some point.

danielfeismann: I see the danger, that someone will use this directly on Kelvin which the get's converted into Celsius. But let's see...

staudtMarius: There is only danger, if someone uses the `.value` method for a squants quantity. If we use the `toKelvin` method, then we will get the correct value.

src/test/scala/edu/ie3/util/quantities/QuantityConversionUtilsSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import squants.thermal.{Celsius, Temperature}
+import tech.units.indriya.quantity.Quantities
+import tech.units.indriya.unit.Units.{CELSIUS, KELVIN}
+
+class QuantityConversionUtilsSpec extends UnitSpec {
+  implicit val doubleTolerance: Double = 1e-10
+  implicit val temperatureTolerance: Temperature = Celsius(1e-10)
+
+  "QuantityConversionUtils" should {
+    "properly convert from ComparableQuantity[Celsius] to squants temperatures and its double values" in {
```
staudtMarius: It is `Temperature` instead of `Celsius`.

staudtMarius (changes requested):
One last thing.

src/test/scala/edu/ie3/util/quantities/QuantityConversionUtilsSpec.scala
danielfeismann: True :)

staudtMarius (approved):
LGTM

PR #1433, opened 2025-06-30
Fixed typo in `HpModel` for model type

danielfeismann:
resolves #1432

PR #1435, opened 2025-06-30
Fix expected values `maxPower` and `energyScaling` for `ProfileLoadMo…

### LINKED ISSUE #1434: Fix  expected values maxPower and energyScaling for ProfileLoadModel Factory

danielfeismann:
…del.Factory`

resolves #1434 

merge #1433 first

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/participant/load/ProfileLoadModel.scala
+++ b/src/main/scala/edu/ie3/simona/model/participant/load/ProfileLoadModel.scala
@@ -25,7 +25,7 @@ import edu.ie3.simona.service.ServiceType.LoadProfileService
 import edu.ie3.simona.service.{Data, ServiceType}
 import edu.ie3.util.scala.quantities.ApparentPower
 import edu.ie3.util.scala.quantities.DefaultQuantities.zeroKW
-import squants.energy.{Energy, Kilowatts}
+import squants.energy.Energy
 import squants.{Dimensionless, Power}
 import edu.ie3.util.scala.quantities.QuantityConversionUtils.PowerConversionSimona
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/ie3/simona/model/participant/load/ProfileLoadModel.scala
```diff
@@ -88,11 +90,7 @@ class ProfileLoadModel(
         case loadFunction: LoadDataFunction =>
           loadFunction.powerSupplier()
       }
-      .getOrElse(
-        throw new CriticalFailureException(
-          s"Expected LoadProfileData, got $receivedData"
-        )
-      )
+      .getOrElse(zeroKW)
```
danielfeismann, line 93: This is not correct here.

The better way would be to fix the calculation of averagePower for the lastTick of simulation, to average with the power value from second last tick in simulation and zeroKW for the last tick in simulation

staudtMarius:
The profile load model should receive all necessary data from the `LoadProfileService`. The issue of missing values for `maxPower` and `energyScaling` will be fixed with #1437.

src/main/scala/edu/ie3/simona/model/participant/load/ProfileLoadModel.scala
staudtMarius: The `ProfileLoadModel` shouldn't do any calculations, since we are using a load profile. This issue is probably realted to the resolution used by the service. This will be resolved with #1444.

danielfeismann: Yes, as discussed before I kept this one open till we have a solution in the other PR. But no need to have a deeper look here, so far.

PR #1437, opened 2025-07-01
Refactored `LoadModel` to fix bugs when using primary data

### LINKED ISSUE #1436: Refactor `LoadModel` to fix bugs when using primary data
There are currently two issues with the load models related to the `modelBehaviour`.

1. The `random` model behaviour can be removed. To use this behaviour one needs to set the load profile of the model to random and choose the `profile` model behaviour in the config.
2. When using the `profile` model behaviour with primary data, we don't have the load profile information which leads to an exception. Therefore, we will use the `FixedLoadModel`, that will be build, when using a load model with primary data.

staudtMarius:
Resolved #1436


sebastian-peter (approved):
LGTM thanks!

PR #1439, opened 2025-07-01
Ambient temperature can cause the next threshold of the ThermalHouse to be the same as the current tick

### LINKED ISSUE #1438: Ambient temperature can cause the next threshold of the ThermalHouse to be the same as the current tick
`ThermalHouse.determineNextThreshold()` should be adapted from
```
else if (
      isInnerTemperatureTooHigh(limitTemperature - temperatureTolerance)
    ) { /* House has more gain than losses */
```
to
```
else if (
      isInnerTemperatureTooHigh(limitTemperature - temperatureTolerance) && qDot > zeroKW
    ) { /* House has more gain than losses AND gain is not caused external (through ambient temperature) */
```

to not calculate any next threshold when temperature rise is caused external.

danielfeismann:
resolves #1438


staudtMarius (approved):
LGTM

PR #1444, opened 2025-07-04
Made powerflow calculation optional.

### LINKED ISSUE #1443: Remove dependence on powerflow resolution
Currently, we use the power flow resolution wherever we need a resolution. But there are cases, where we want to run SIMONA with a resolution, that deviate from the power flow resolution.

staudtMarius:
Resolves #1443


src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
```diff
@@ -71,20 +70,28 @@ object GridAgent extends DBFSAlgorithm with DCMAlgorithm {
...
+
+    // this determines the agents regular time bin it wants to be triggered e.g. one hour
+    // if no resolution is given, we use the end time to prevent powerflow calculation
+    val resolution: Long = cfg.powerflow match {
+      case Some(value) =>
+        value.resolution.toSeconds
+      case None =>
+        // powerflow is only performed after the simulation ended
+        simEndTime.toTick + 1
+    }
```
danielfeismann: Why do we perform a powerflow also in the case we don't want to perform a powerflow?

danielfeismann (changes requested):

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
staudtMarius: We don't perform a powerflow, since the tick of the powerflow is one tick after the last tick of the simulation. If you have any suggestion for improvements, please let me know.

danielfeismann: That's my point. It looks a bit odd here. How about making resolution Option[Long]?

danielfeismann (approved):
LGTM, but @sebastian-peter should also have a look on this.

docs/readthedocs/config.md
```diff
@@ -310,6 +310,9 @@ are described in the documentation of [voltage limits](models/voltage_limits).
 
 ## Power flow configuration 
 
+The powerflow configuration in SIMONA is an optional configuration. If no configuration is present, SIMONA will skip all
```
danielfeismann, line 313: may we add a comment in 'vn_simona.conf' for this?

danielfeismann (changes requested):

src/main/scala/edu/ie3/simona/service/load/LoadProfileService.scala
```diff
@@ -47,20 +48,20 @@ object LoadProfileService extends SimonaService {
...
-    *   The resolution of this simulation.
     */
   final case class LoadProfileInitializedStateData(
       loadProfileStore: LoadProfileStore,
       profileToRefs: Map[LoadProfile, Seq[ActorRef[ParticipantAgent.Request]]] =
         Map.empty,
-      nextActivationTick: Long,
+      maybeNextActivationTick: Option[Long],
+      activationTicks: SortedDistinctSeq[Long] = SortedDistinctSeq.empty,
       simulationStartTime: ZonedDateTime,
```
sebastian-peter: Why give up on nextTick + resolution? It fits the DES concept quite nicely and takes less memory.

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
```diff
@@ -171,25 +183,21 @@ object GridAgent extends DBFSAlgorithm with DCMAlgorithm {
...
-        constantData.resolution,
-        Some(unlockKey),
-      )
+      constantData.resolution.foreach { resolution =>
+        constantData.environmentRefs.scheduler ! ScheduleActivation(
+          ctx.self,
+          resolution,
+          Some(unlockKey),
+        )
+      }
```
sebastian-peter: Wouldn't it be easier if we just don't schedule ourselves, if powerflow is not needed? Or is there any other task that GridAgent can perform if it doesn't calculate pf? Recognizing/solving congestions should not be possible without it, right?

sebastian-peter (commented):
Love the idea, just some questions about the implementation from my side

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
staudtMarius: This message is only sent, if we have a powerflow resolution. But I will adapt this part to make it clearer.

src/main/scala/edu/ie3/simona/service/load/LoadProfileService.scala
staudtMarius: The problem was the resolution part. We used the powerflow resolution, which caused some problems. Since the load profiles are sources, we can ask them for the next tick. I will try to reduce the memory consumption.

sebastian-peter: A fixed resolution per load profile would have been fine too, but this works too I guess

staudtMarius: We could adapt the LoadProfileTimeSeries to provide this information.

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
sebastian-peter: My point is, `resolution` in `GridAgentConstantData` does probably not have to be optional, because if it's `None`, GridAgent won't be activated again. I thought about stopping the GA if there's no pf, but that would probably complicate things as well. So maybe just picking any default value for resolution would work

staudtMarius: We could use `Long.MaxValue` instead.

docs/readthedocs/config.md
staudtMarius: I added a coment in config. But I would like to keep this info also in the documentation.

danielfeismann: Yes, both please.

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
```diff
@@ -191,14 +192,21 @@ object GridAgent extends DBFSAlgorithm with DCMAlgorithm {
         SimonaActorNaming.actorName(ctx.self),
       )
 
-      constantData.resolution.foreach { resolution =>
-        constantData.environmentRefs.scheduler ! ScheduleActivation(
+      val resolution = constantData.resolution
+
+      val messageToScheduler = if (resolution == Long.MaxValue) {
+        unlockKey.unlock()
+        Completion(ctx.self)
```
sebastian-peter: I think this doesn't work, since we have not been activated before

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
```diff
@@ -76,8 +76,9 @@ object GridAgent extends DBFSAlgorithm with DCMAlgorithm {
     val simEndTime = cfg.time.simEndTime
 
     // this determines the agents regular time bin it wants to be triggered e.g. one hour
-    // if no resolution is given, we use the end time to prevent powerflow calculation
-    val resolution = cfg.powerflow.map(_.resolution.toSeconds)
+    // if no resolution is given, we use the maximal long value, to prevent unwanted power flow calculation
```
sebastian-peter:
```suggestion
    // if no resolution is given, we use the maximal long value as a placeholder, as no future activation should take place
```

src/main/scala/edu/ie3/simona/agent/grid/GridAgentData.scala
```diff
@@ -43,15 +43,15 @@ object GridAgentData {
...
+    *   Environment actor refs.
     * @param simonaConfig
-    *   config
+    *   Config.
     * @param listener
-    *   listeners
+    *   Listeners.
     * @param resolution
-    *   of the simulation
+    *   Of the simulation.
```
sebastian-peter: ... Maybe enhancing this a bit would be great, while you're at it. This doesn't look too nice.

sebastian-peter (changes requested):

src/main/scala/edu/ie3/simona/agent/grid/GridAgent.scala
staudtMarius: Do you mean the unlock or the message?

sebastian-peter: The message. The unlock is still necessary

staudtMarius: Ok. I will adapt this part.

sebastian-peter (dismissed):
Thanks for the adaptations, looks good!

input/samples/vn_simona/vn_simona.conf
```diff
@@ -210,6 +210,7 @@ simona.gridConfig.refSystems = [
 ##################################################################
 # Power Flow Configuration
 ##################################################################
+# The powerflow configuration in SIMONA is an optional configuration. If no configuration is present, SIMONA will skip all powerflow calculations.
 simona.powerflow.maxSweepPowerDeviation = 1E-5 // the maximum allowed deviation in power between two sweeps, before overall convergence is assumed
 simona.powerflow.newtonraphson.epsilon = [1E-12]
 simona.powerflow.newtonraphson.iterations = 50
```
danielfeismann: Does the simona.control.transformer part belong to the section of power flow config? If not it should be moved to not confuse one.

danielfeismann (changes requested):

danielfeismann (approved):
LGTM :)

PR #1476, opened 2025-07-30
Rename thermalStorage to heatStorage

### LINKED ISSUE #1475: Rename thermalStorage to heatStorage
to align code and prepare for introduction of DomesticHotWaterStorage

danielfeismann:
resolves #1475 

merge first
- #1485

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Made powerflow parameters in config options [#1443](https://github.com/ie3-institute/simona/issues/1443)
 - Provide abstraction for flex options in system participant and EM [#1456](https://github.com/ie3-institute/simona/issues/1456)
 - Rename `MinMaxFlexOptions` to `PowerLimitFlexOptions` [#1457](https://github.com/ie3-institute/simona/issues/1457)
+- Rename 'thermalStorage' to 'heatStorage' [#1475](https://github.com/ie3-institute/simona/issues/1475)
 
 ### Fixed
 - Fixes in Documentation, ScalaDocs, Code Style and more [#1397](https://github.com/ie3-institute/simona/issues/1397)
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -51,11 +51,11 @@ import java.util.UUID
   * @param ethCapa
   *   heat energy storage capability of thermal house, usually in [kWh/K]
   * @param targetTemperature
-  *   Target room temperature [K]
+  *   Target room temperature [°C]
   * @param lowerBoundaryTemperature
-  *   Lower temperature boundary [K]
+  *   Lower temperature boundary [°C]
   * @param upperBoundaryTemperature
-  *   Upper boundary temperature [K]
+  *   Upper boundary temperature [°C]
   */
 final case class ThermalHouse(
     uuid: UUID,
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Made powerflow parameters in config options [#1443](https://github.com/ie3-institute/simona/issues/1443)
 - Provide abstraction for flex options in system participant and EM [#1456](https://github.com/ie3-institute/simona/issues/1456)
 - Rename `MinMaxFlexOptions` to `PowerLimitFlexOptions` [#1457](https://github.com/ie3-institute/simona/issues/1457)
+- Rename 'thermalStorage' to 'heatStorage' [#1475](https://github.com/ie3-institute/simona/issues/1475)
```
staudtMarius, line 34: Is this style intended? Is doesn't match our usual style.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -51,11 +51,11 @@ import java.util.UUID
...
   *   heat energy storage capability of thermal house, usually in [kWh/K]
   * @param targetTemperature
-  *   Target room temperature [K]
+  *   Target room temperature [°C]
   * @param lowerBoundaryTemperature
-  *   Lower temperature boundary [K]
+  *   Lower temperature boundary [°C]
   * @param upperBoundaryTemperature
-  *   Upper boundary temperature [K]
+  *   Upper boundary temperature [°C]
```
staudtMarius: In the `apply` method, we are using Kelvin. Which unit should be used?

staudtMarius (changes requested):
LGTM. I have just two small questions.

sebastian-peter (commented):
Stupid question: Why rename it at all? The term thermal storage seems like it includes storing "coldness" as well, which is something you want to do too?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: Good catch! PSDM is in Celsius, the apply should be as well imho. But since this is now more than just a ScalaDoc-Error as I first thought. Thus, I will move this part into a separate PR.

CHANGELOG.md
danielfeismann: Yes, I choosed this style since ThermalStorage as class will remain and it only refers to the variables. Do you have a suggestion, I'll be happy to include it.

danielfeismann:
> Stupid question: Why rename it at all? The term thermal storage seems like it includes storing "coldness" as well, which is something you want to do too?

No. The simple reason for this is that, after introducing the calculation of hot water consumption, there are two storage tanks, making it easier to distinguish between the two heat storage tanks.

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
```diff
@@ -102,19 +102,19 @@ class ThermalGridWithHouseAndStorageSpec
                 ),
                 Some(
                   ThermalStorageState(
-                    storageTick,
-                    storedEnergy,
+                    tickHeatStorage,
```
staudtMarius, line 105: Why did you use `tickHeatStorage` instead of `heatStorageTick`?

staudtMarius (changes requested):
Just one more question.

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
danielfeismann: To have the parameter of interest first. Like in `storedEnergyHeatStorage`. One could now argue that it should also be tickHouse. However, I think it's fine not to be one hundred percent consistent here.

staudtMarius (approved):
LGTM.

PR #1485, opened 2025-07-31
Align temperature scale of parameter for `ThermalInputs`

### LINKED ISSUE #1484: Align used temperature scale for thermal input parameter
In the `apply` method, we are using Kelvin. Which unit should be used?

_Originally posted by @staudtMarius in https://github.com/ie3-institute/simona/pull/1476#discussion_r2245466563_

danielfeismann:
resolves #1484

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -27,7 +27,7 @@ import edu.ie3.util.scala.quantities.DefaultQuantities._
 import edu.ie3.util.scala.quantities.SquantsUtils.RichThermalCapacity
 import edu.ie3.util.scala.quantities.{ThermalConductance, WattsPerKelvin}
 import squants.energy.KilowattHours
-import squants.thermal.{Kelvin, ThermalCapacity}
+import squants.thermal.{Celsius, Kelvin, ThermalCapacity}
 import squants.time.Seconds
 import squants.{Energy, Power, Temperature, Time}
 import tech.units.indriya.unit.Units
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -334,14 +334,14 @@ object ThermalHouse {
         .getValue
         .doubleValue
     ) / Kelvin(1d),
-    Kelvin(
-      input.getTargetTemperature.to(Units.KELVIN).getValue.doubleValue
+    Celsius(
+      input.getTargetTemperature.to(Units.CELSIUS).getValue.doubleValue
```
staudtMarius: We could use the existing convertion util `TemperatureConversionSimona` here.

staudtMarius (changes requested):
A suggestion for a small code improvement.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: right

staudtMarius (approved):
LGTM.

sebastian-peter:
@danielfeismann @staudtMarius The point of `Temperature` (and all other quantities) is to abstract away from concrete units. So, I don't think it's necessary at all to specify the unit of measurement of `Temperature` parameters. Or am I wrong? We don't tell people to provide kW when we ask for a `Power` as well, right?

staudtMarius:
@sebastian-peter We should probably open another issue to remove this part of the scaladoc.

danielfeismann:
@sebastian-peter thats right. I'm fine with removing it from scalaDoc (at both places). But before it was noted that we expect it in Kelvin, which was right then, but differ from what we write in PSDM. I'm sure this was hard to get for some person not familiar with simona. Resolving this possibility for confusion was my intention here.

PR #1486, opened 2025-08-04
Simplify flex option result configuration

### LINKED ISSUE #1111: Simplify flex option result configuration
Currently, flex results have to be turned on in for types of system participants in `simona.output.participant`, but also setting `simona.output.flex = true`. The latter should not be necessary, since any `flexResult = true` for participant output settings should suffice to turn flex outputs on. Maybe `ConfigUtil` should be enhanced to deal with flex.

danielfeismann:
resolves #1111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Made powerflow parameters in config options [#1443](https://github.com/ie3-institute/simona/issues/1443)
 - Provide abstraction for flex options in system participant and EM [#1456](https://github.com/ie3-institute/simona/issues/1456)
 - Rename `MinMaxFlexOptions` to `PowerLimitFlexOptions` [#1457](https://github.com/ie3-institute/simona/issues/1457)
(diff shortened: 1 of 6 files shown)


sebastian-peter:
Thanks a lot for tackling this. I think the status quo is very problematic for obvious reasons.

I would even go a step further than this, tbh. I think `output.flex` purely exists for historic reasons. The only justification for its existence that I could come up with would be that there's global configuration for grid outputs (nodes, lines, etc.), and `output.flex` is modeled with that example in mind.

This is not very consistent, as flex output configuration is related to system participants and EMs, and these assets have always been configured individually (with an option for global configuration via default parameter).

Plus, I think there's no way to justify needing two configuration parameters to enable one feature. And btw, I think your PR makes `output.flex` useless anyway, because it doesn't affect outputs anymore, whether set to true or false.

tl;dr I'd suggest removing `output.flex`

danielfeismann:
Good point, this was mainly what motivated me to tackle this. Before this one has to set output.flex global to get any flexResult, which easily can be missed.

Please have a look, I'm not sure if I got everything with the new config scheme.

sebastian-peter (approved):
Looks good, thanks. Made some tiny additions, let me know what you think.

PR #1489, opened 2025-08-07
Harmonisation of Transformer Test Data

### LINKED ISSUE #163: Harmonization of Transformer Test Data
`TransformerInputTestData.scala`, `ThreeWindingTestData.scala` and `Transformer3wTestData` seems to contain similar input data for transformer tests. Maybe these could be merged?

This came up at PR #155 where changes on the susceptance value bM doesn't effect any test => negative bM should be correct, but tests also accept positive bM since the data in ThreeWindingTestData doesn't end in the test which check the results for different susceptances from thin input...)

danielfeismann:
resolves #163


sebastian-peter (approved):
Looks good, thanks!

Just for the sake of completeness: The nominal voltage of node c is different now, but this is totally fine since it doesn't matter to `GridAgentSetupSpec`.

danielfeismann:
> Looks good, thanks!
> 
> Just for the sake of completeness: The nominal voltage of node c is different now, but this is totally fine since it doesn't matter to `GridAgentSetupSpec`.

Yes, but it was easier this way than fixing the voltage level in the tests of the other node c...

PR #1496, opened 2025-08-22
Removed some units from scalaDoc within `ThermalHouse`

### LINKED ISSUE #1495: Remove units for temperature in scalaDoc
@danielfeismann @staudtMarius The point of `Temperature` (and all other quantities) is to abstract away from concrete units. So, I don't think it's necessary at all to specify the unit of measurement of `Temperature` parameters. Or am I wrong? We don't tell people to provide kW when we ask for a `Power` as well, right?

_Originally posted by @sebastian-peter in https://github.com/ie3-institute/simona/issues/1485#issuecomment-3167907102_

danielfeismann:
resolves #1495


sebastian-peter (approved):
LGTM thanks!

PR #1501, opened 2025-08-27
Introducing optimizing energy managament strategy

### LINKED ISSUE #1500: Introduce optimizing EM strategy

sebastian-peter:
Resolves #1500

This PR also introduces storage flex options and power flex options for  mathematical programming, because otherwise there would be nothing to test the flex strategy with.


src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  override def determineFlexControl(
+      flexOptions: Iterable[
+        (? <: AssetInput, MathFlexOptions[?, ? <: OperationVars])
+      ],
+      target: Power,
+      currentTick: Long,
+  ): Iterable[(UUID, Power)] = {
+
+    implicit val model: MPModel = MPModel(SolverLib.oJSolver)
```
staudtMarius: Could be changed to `given model: MPModel`.

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,338 @@
...
+    * @param model
+    *   The optimization model to use.
+    * @tparam SV
+    *   The type of state variables.
+    * @tparam OV
+    *   The type of operation variables.
+    * @return
+    *   A container that holds all state and operation variables.
+    */
+  def addAssetConstraints[SV, OV <: OperationVars](
```
staudtMarius, line 132: In my opinion, this method should be named `buildAssetContrains`, because the method builds and returns the constraints instead of adding them to another object.

staudtMarius (changes requested):
I have two small suggestions.

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
sebastian-peter: It's a bit more complicated than that. The method creates and adds the constraints to the model, and then returns the _variables used_ for creating the constraints. So I'd say the name is okay. One could find a longer name that is more descriptive, but we have tried that before and it was never really beneficial, imho (up until now, we've replaced them with shorter names at many places).

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
```diff
@@ -0,0 +1,415 @@
...
+
+          // charging until full, with maximum power
+          val power1 = pMax
+          val duration1 = (eStorage - currentEnergy) / (power1 * etaCharging)
+          // discharging three quarters, with half power
+          val power2 = -pMax / 2
+          val time2 = eStorage * 0.75 / (-power2 * 1 / etaDischarging)
+          // charging until half, with quarter power
+          val power3 = pMax / 4
+          val time3 = eStorage / 4 / (power3 * etaCharging)
```
danielfeismann, line 61:
```suggestion
          val time3 = eStorage / 2 / (power3 * etaCharging)
```

or "charging until one quarter" in the comment above, or?

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,338 @@
...
+import squants.{Power, Time}
+
+import java.util.UUID
+
+/** Flex strategy that optimizes over a fixed amount of time steps into the
+  * future. Works with [[MathFlexOptions]], which describe constraints on
+  * present and future behavior of a participant.
+  *
+  * @param sampleTime
+  *   The amount of time in between steps.
```
danielfeismann:
```suggestion
  *   The amount of time between the steps.
```

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,338 @@
...
+  * @param predictionHorizon
+  *   The amount of time that is considered in the future, i.e. the last step is
```
danielfeismann:
```suggestion
  *   The amount of time that is predicted into the future, i.e. the last step is
```

as i would interpret a "predictionHorizon"

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,338 @@
...
+      flexOptions: MathFlexOptions[?, ? <: OperationVars],
+  ): MathFlexOptions[?, ? <: OperationVars] = flexOptions
+}
+
+object OptimizedFlexStrat {
+
+  /** Threshold for the error of soft constraints after optimization. Every soft
+    * constraint error should stay below this threshold.
+    */
+  private val softConstraintThreshold: Double = 1e-3
```
danielfeismann, line 111: Do you need this because the numerical solution might differ slidely and thus breaks the real constraint a bit, or is there another reason why to have soft constraints? This sounds a bit like the threshold handling when charging / discharging a storage...

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
```diff
@@ -0,0 +1,233 @@
...
+      * @param currentEnergy
+      *   The currently stored energy.
+      * @param eStorage
+      *   The storage capacity.
+      * @param pMax
+      *   The maximum charging and discharging power.
+      * @param etaCharging
+      *   The charging efficiency.
+      * @param etaDischarging
+      *   The discharging efficiency.
```
danielfeismann, line 128: Since PSDM atm not distinguish between eta for charging and discharging, I would propose to have only eta here.

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
```diff
@@ -0,0 +1,233 @@
...
+        formerState: StorageStateVars,
+        op: StorageOperationVars,
+        tick: Long,
+    )(using model: MPModel): StorageStateVars = {
+
+      val storedEnergy = MPFloatVar("storedEnergy", 0, eStorage.toKilowattHours)
+      val timeInHours = (tick - formerState.tick) / 3600
+
+      model.add(
+        storedEnergy := formerState.storedEnergy + (op.p - op.pAbs * (1 - eta.toEach)) * timeInHours
```
danielfeismann, line 106: Can't this be done with squants, I see some danger of mixing wrong units here

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
```diff
@@ -0,0 +1,233 @@
...
+        eStorage: Energy,
+        pMax: Power,
+        etaCharging: Dimensionless,
+        etaDischarging: Dimensionless,
+    ): StorageMathFlexOptions = {
+
+      val etaCh = etaCharging.toEach
+      val etaDis = etaDischarging.toEach
+
+      val etaAvg = (2 * etaCh * etaDis) / (1 + etaCh * etaDis)
```
danielfeismann, line 143: Could you help me what's the purpose here?

src/test/scala/edu/ie3/simona/model/em/OptimizedFlexStratIT.scala
```diff
@@ -0,0 +1,266 @@
...
+          batVars.operationVars(0).pVal should approximate(0d)
+          batVars.states(1).energyVal should approximate(0d)
+
+          // 6 kW of feed-in to compensate, 2 kW remains
+          batVars.operationVars(1).pVal should approximate(4d)
+          batVars.states(2).energyVal should approximate(3.2d)
+
+          // 8 kW of feed-in to compensate, 2 kW remains
+          batVars.operationVars(2).pVal should approximate(6d)
+          batVars.states(3).energyVal should approximate(8d)
```
danielfeismann, line 211: Could you help me why it is 8 kWh? (assuming one hour steps, 6 kW * 0.8 + 3.2 from previous step isn't 8.

danielfeismann: Note to self: have to check the rest here.

src/test/scala/edu/ie3/simona/test/common/MathFlexTestLike.scala
```diff
@@ -0,0 +1,58 @@
...
+      state.p.value.value
+
+  final case class EnergyConversionFactor(factor: Double)
+
+  object EnergyConversionFactor {
+    def apply(
+        regularChargingEta: Dimensionless,
+        adaptedEta: Dimensionless,
+    ): EnergyConversionFactor =
+      new EnergyConversionFactor(regularChargingEta.toEach / adaptedEta.toEach)
```
danielfeismann: Do you need the conversion into double, via toEach, here?

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
```diff
@@ -0,0 +1,415 @@
...
+        container.states(3).energyVal should (be >= 0d and be <= 20d)
+
+        // possibly discharging, now we should have reached 0 kWh
+        container.operationVars(3).pVal should be <= 0d
+        container.states(4).energyVal should approximate(0d)
+
+        // we should've discharged 20 kWh minus 4 kWh losses
+        val totalDischarged =
+          container.operationVars(2).pVal + container.operationVars(3).pVal
+        totalDischarged should approximate(-16d)
```
danielfeismann, line 365: Maybe total discharged is a bit misleading. Since the discharge is 20 but the netto output is only 16.

danielfeismann (changes requested):
Just some comments but mainly questions that came up from my side. Maybe we can have a short talk on the linear programming approach which may cause some of my questions.

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
sebastian-peter: No, because we charged to full, then discharged three quarters, thus we're at one quarter and need one more quarter to reach half full.

danielfeismann: Makes sense, didn't get that it these steps are seen as a sequence.

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
sebastian-peter: `in between` seems [fine to me](https://www.merriam-webster.com/dictionary/in-between)?

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
sebastian-peter: I'm using soft constraints here because in pure LP, there is no way to determine the absolute value of a variable (there is no max function etc.). Soft constraint means that I'm adding a portion of the variable to the objective, and here this means adding a part of the absolute-value-variable to the objective in order to keep it as close to the actual absolute as possible. 
In the end, this could mean though that the value I want (the exact absolute) is not achieved (which is probably because the penalty of the soft constraint is too low). But as long as the error (i.e. the difference between expected and actual value) is below this threshold, I'm fine with the result.

Let me know if I can clarify more here... I think my explanation here is not that good

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
sebastian-peter: I'm kind of thinking in the opposite direction... I got my storage math model to work with split efficiencies, which was a little bit more complicated than just a single efficiency. So I'd rather keep it this way and maybe add optional parameters to PSDM.

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
sebastian-peter: Nope, the mathematical model does not allow for units

src/main/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModel.scala
sebastian-peter: I created a storage model that purely works in linear programming, but in order to do so, I had to adapt the storage model a bit. The same loss is subtracted for charging and discharging operations, which means we do not need to distinguish between charging and discharging, which means we don't need binary variables (which would make the problem mixed-integer). But in order to do so, we need to calculate an average eta and also adapt the stored energy. In the end though, the regular and adapted model work in the same way and can be transformed into each other.

I sort-of wrote an abstract on this, it's probably easiest if I show it to you.

src/test/scala/edu/ie3/simona/model/em/OptimizedFlexStratIT.scala
sebastian-peter: `6 * 0.8 + 3.2 = 8` works out for me...

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
sebastian-peter: I see what you mean... Do you have a suggestion for a different naming?

src/test/scala/edu/ie3/simona/test/common/MathFlexTestLike.scala
sebastian-peter: No, guess it works without. Thanks for the hint!

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
danielfeismann: mh `outputDischarging` or `nettoDischarged`

src/test/scala/edu/ie3/simona/model/em/OptimizedFlexStratIT.scala
danielfeismann: Whatever I calculated yesterday, yes that's working :D

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
danielfeismann: I think it is okay but between works as well and sounds nicer to me :) And, [Between, on its own, is often the correct choice](https://www.grammarly.com/blog/commonly-confused-words/in-between-or-inbetween/)

sebastian-peter: Ok, good point

src/test/scala/edu/ie3/simona/model/participant/storage/StorageMathFlexModelSpec.scala
sebastian-peter: I used `input`/`output` now, since `net` is oftentimes construed as being less than `gross`, which is not the case for charging here

src/main/scala/edu/ie3/simona/model/em/OptimizedFlexStrat.scala
danielfeismann: Well, I think my confusion was more caused by the lack of how and why integrating the LP. There might be options for making this description clearer, but atm I don't have concrete one. So let's keep it as it is.

danielfeismann (approved):
Thanks for refering and explaining to my questions. This looks fine to me :)

staudtMarius (approved):

PR #1517, opened 2025-09-04
Unifying coordinate class

### LINKED ISSUE #1516: Unify coordinate class
In dealing with weather, we currently have `WeatherService.Coordinate` and `WeatherSource.AgentCoordinates`, which can be unified

sebastian-peter:
Resolves #1516

### DIFF
--- a/src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
+++ b/src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
@@ -24,11 +24,8 @@ import edu.ie3.simona.ontology.messages.ServiceMessage.{
   SecondaryServiceRegistrationMessage,
   ServiceRegistrationMessage,
 }
-import edu.ie3.simona.service.weather.WeatherSource.{
-  AgentCoordinates,
-  WeightedCoordinates,
-}
-import edu.ie3.simona.util.SimonaConstants
+import edu.ie3.simona.service.weather.WeatherSource.WeightedCoordinates
+import edu.ie3.simona.util.{Coordinate, SimonaConstants}
 import edu.ie3.simona.util.TickUtil.RichZonedDateTime
 import edu.ie3.util.scala.collection.immutable.SortedDistinctSeq
 import org.apache.pekko.actor.typed.ActorRef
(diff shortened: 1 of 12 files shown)


src/main/scala/edu/ie3/simona/service/weather/WeatherSource.scala
```diff
@@ -242,15 +239,15 @@ trait WeatherSource {
     *
     * @param tick
     *   Simulation date in question
-    * @param agentToWeightedCoordinates
-    *   The coordinates in question
+    * @param coordinateWeights
+    *   The coordinates and respective weights in question
```
staudtMarius, line 242: Missing `.` at the end. Also for the other descriptions.

src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
```diff
@@ -24,11 +24,8 @@ import edu.ie3.simona.ontology.messages.ServiceMessage.{
   SecondaryServiceRegistrationMessage,
   ServiceRegistrationMessage,
 }
-import edu.ie3.simona.service.weather.WeatherSource.{
-  AgentCoordinates,
-  WeightedCoordinates,
-}
-import edu.ie3.simona.util.SimonaConstants
+import edu.ie3.simona.service.weather.WeatherSource.WeightedCoordinates
```
staudtMarius, line 27: Does it makes sense to move the `WeightedCoordinates` into the `Coordinate` class?

staudtMarius (changes requested):
Looks already good. I have just two small points.

src/main/scala/edu/ie3/simona/service/weather/WeatherSource.scala
sebastian-peter: Fixed all scaladoc in WeatherSource, this should be enough for now

src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
sebastian-peter: I don't think so, as the usage should be limited to the weather service.

src/main/scala/edu/ie3/simona/service/weather/WeatherSource.scala
staudtMarius: There is one last missing `.` in line 242.

sebastian-peter: Ok, thanks for double checking :)

staudtMarius (approved):
LGTM.

PR #1537, opened 2025-09-12
Introducing forecast calculation to `PvModel` and `WecModel`

### LINKED ISSUE #1513: Implement power forecast for PV and WEC

sebastian-peter:
Resolves #1513 

Merge first:
- #1501


src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
```diff
@@ -114,9 +123,17 @@ class PvModel private (
   override def determineOperatingPoint(
       state: PvState
   ): (ActivePowerOperatingPoint, Option[Long]) = {
+    val (_, radiationData) = state.radiationData
+      .maxBefore(state.tick + 1)
```
staudtMarius, line 127: Is this the only way to get the value for the current tick?

src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
```diff
@@ -246,23 +263,98 @@ class PvModel private (
...
     *   The current tick.
     * @param dateTime
-    *   The date and time of the <b>ending</b> of time frame to calculate.
-    * @param diffIrradiance
-    *   The diffuse solar irradiance on a horizontal surface.
-    * @param dirIrradiance
-    *   The direct solar irradiance on a horizontal surface.
+    *   The date and time corresponding to the current tick.
+    * @param radiationData
+    *   A map of tick to corresponding radiation data. For regular calculations
```
staudtMarius, line 273: `ticks`?

src/main/scala/edu/ie3/simona/model/participant/WecModel.scala
```diff
@@ -86,31 +91,39 @@ class WecModel private (
...
-      )
+      .map(newData => state.copy(weatherData = newData))
       .getOrElse(state)
 
   override def determineOperatingPoint(
       state: WecState
   ): (ActivePowerOperatingPoint, Option[Long]) = {
-    val betzCoefficient = determineBetzCoefficient(state.windVelocity)
+    val (_, weatherData) = state.weatherData
+      .maxBefore(state.tick + 1)
```
staudtMarius, line 109: Same as in `PvModel`.

src/main/scala/edu/ie3/simona/model/participant/WecModel.scala
```diff
@@ -229,21 +242,95 @@ object WecModel {
     */
   private val molarMassAir = Kilograms(0.0289647d)
 
-  /** Holds all relevant data for wec model calculation
+  /** Holds all relevant data for WEC model calculation.
+    *
+    * @param tick
+    *   The current tick.
+    * @param weatherData
+    *   A map of tick to corresponding weather data. For regular calculations of
```
staudtMarius, line 250: Same as in `PvModel`.

staudtMarius (changes requested):
Look good to me. I just have some small remarks.

src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
sebastian-peter: What's the problem with singular here?

src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
sebastian-peter: There is `minAfter`, which considers the smallest key _larger or equal to_ the key, and there's `maxBefore`, which looks for the largest key _less than_ the given key. That's the way `SortedMap` works in scala and I don't see any other simple way.

sebastian-peter: One could do `rangeTo(state.tick).lastOption`, do you like that more? In the end, both `rangeTo` and `maxBefore` are built upon `rangeUntil`, so it doesn't really matter after all, imho.

src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
staudtMarius: You could have data for multiple ticks.

sebastian-peter: Yes, and still it would be a map from tick to data?

sebastian-peter: Actually, since the map is bijective, it'd make more sense to keep it singular... A tick only maps to one data object, and data objects are mapped to by only one tick each

src/main/scala/edu/ie3/simona/model/participant/PvModel.scala
staudtMarius: Ok. I think we should keep it this way.

staudtMarius (approved):
LGTM.

PR #1540, opened 2025-09-17
Cleaning up dependencies

### LINKED ISSUE #1539: Clean up dependencies

sebastian-peter:
Resolves #1539

- `org.apache.commons:commons-csv`: Only used in tests -> `testImplementation`, results in smaller jar
- `org.locationtech.jts.io:jts-io-common`: Not used at all
- `org.apache.commons:commons-math3`: Not used at all
- `org.apache.poi:poi-ooxml:`: We only use it for `FileNameUtils`, thus `commons-io:commons-io` suffices

With a little bit of grit, probably more could be done:
- `jgrapht`: Only used for checking connectivity in `GridModel`, which is already implemented in PSDM  (`ConnectorValidationUtils#checkConnectivity`)
- `breeze`: Only used in `DbfsAlgorithm`, not sure if really required
- `jts`: Only used for its `Point` class
- `scopt`: We don't do much argument parsing besides `--config` anyways

Also, one could comb through our own dependencies and do the same there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed CFF-Version [#1464](https://github.com/ie3-institute/simona/issues/1464)
 - Fixed handleInput of `ProfileLoadModel` [#1441](https://github.com/ie3-institute/simona/issues/1441)
 - Fixed compiler warnings related to import and scala3 syntax [#1383](https://github.com/ie3-institute/simona/issues/1383)
+- Cleaned up dependencies [#1539](https://github.com/ie3-institute/simona/issues/1539)
 
 ### Removed
 - Removed unused classes and methods related to pekko classic actors [#1389](https://github.com/ie3-institute/simona/issues/1389)
--- a/build.gradle
+++ b/build.gradle
@@ -125,6 +125,9 @@ dependencies {
   testImplementation "com.dimafeng:testcontainers-scala-postgresql_${scalaVersion}:${testContainerVersion}"
   testImplementation "com.dimafeng:testcontainers-scala-kafka_${scalaVersion}:${testContainerVersion}"
 
+  // test-related csv
+  testImplementation 'org.apache.commons:commons-csv:1.14.1'
+
   /* --- Scala libs --- */
   /* CORE Scala */
   implementation "org.scala-lang:scala3-library_3:${scalaBinaryVersion}"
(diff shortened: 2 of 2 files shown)


staudtMarius:
@sebastian-peter 

- If you want to remove `jgrapht`, we have to do some code adaption, since we need it currently to use the `SubGridTopologyGraph` in SIMONA.
- `jts` has the same problem as `jgrapht`. If we remove the dependency, the compiler has a problem to identify the class on the classpath. 

Maybe we can fix this with some compiler option.

- `Breeze` is currently needed for the `Complex` and `DenseMatrix` classse. It could be removed, after we have done some changes in the `powerflow`.

- `scopt` is also used to read in the external address for `mosaik`. But we could write a pure scala implementation (see: https://www.baeldung.com/scala/read-command-line-arguments), since we only need the `config` in SIMONA. The other arguments could be handled in the external simulation.

staudtMarius (changes requested):
Looks good to me. I added some comments regarding the other dependencies you mentioned.

@sebastian-peter 

- If you want to remove `jgrapht`, we have to do some code adaption, since we need it currently to use the `SubGridTopologyGraph` in SIMONA.
- `jts` has the same problem as `jgrapht`. If we remove the dependency, the compiler has a problem to identify the class on the classpath. 

Maybe we can fix this with some compiler option.

- `Breeze` is currently needed for the `Complex` and `DenseMatrix` classse. It could be removed, after we have done some changes in the `powerflow`.

- `scopt` is also used to read in the external address for `mosaik`. But we could write a pure scala implementation (see: https://www.baeldung.com/scala/read-command-line-arguments), since we only need the `config` in SIMONA. The other arguments could be handled in the external simulation.

What do you mean with `our own dependencies`?

sebastian-peter:
@staudtMarius: With "own dependencies" I mean our own linked libraries, such as PSDM and PSU. And I think this is also the crux of the matter. _If_ we want to clean up more, we should start in our own dependencies, since removing the libraries just in SIMONA does not do anything - a lot of times, they are still included through them.

But, I think a crucial point here is that we should be careful with investing too much time here. I don't know how you see it, but to me, it's not a _major_ goal to me to reduce dependencies, rather a _nice-to-have_.

staudtMarius:
@sebastian-peter I think we should try to replace or remove dependencies where we only used a few classes or methods. But it can be done step by step, e.g. when we are working on the relevant part in the PSDM/PSU anyway.

danielfeismann (approved):
LGTM, but I agree we should be careful with investing too much time here. So remove unused ones and go with that.

staudtMarius (approved):
LGTM.

PR #1548, opened 2025-09-23
Prepare ThermalHouse for changes by including tap water

### LINKED ISSUE #1547: Prepare ThermalHouse for changes by including tap water

danielfeismann:
resolves #1547

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -473,7 +473,7 @@ final case class ThermalHouse(
 
 object ThermalHouse {
   protected def temperatureTolerance: Temperature = Kelvin(0.01d)
-
+  
   /** Temperature values are set constant and based on VDI 2067 Blatt 12.
     */
   protected val lowerTemperatureTapWater: Temperature = Celsius(10d)
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -472,20 +472,20 @@ final case class ThermalHouse(
 }
 
 object ThermalHouse {
-  protected def temperatureTolerance: Temperature = Kelvin(0.01d)
-  
+  private def temperatureTolerance: Temperature = Kelvin(0.01d)
+
   /** Temperature values are set constant and based on VDI 2067 Blatt 12.
     */
-  protected val lowerTemperatureTapWater: Temperature = Celsius(10d)
-  protected val upperTemperatureTapWater: Temperature = Celsius(55d)
+  private val lowerTemperatureTapWater: Temperature = Celsius(10d)
+  private val upperTemperatureTapWater: Temperature = Celsius(55d)
 
   /** Volume values are based on VDI 2067 Blatt 12. Time series relative are
     * based on DIN EN 12831-3 Table B.2 for single family houses and for flats.
     */
-  protected val waterDemandVolumePerPersonYear =
+  private val waterDemandVolumePerPersonYear =
     // Shower and Bath + bathroom sink + dish washing per hand (also dishwasher in the building)
     Litres(8600 + 4200 + 300)
-  protected val waterVolumeRelativeHouse = Map(
+  private val waterVolumeRelativeHouse = Map(
     0 -> 0.018,
     1 -> 0.01,
     2 -> 0.006,
@@ -511,7 +511,7 @@ object ThermalHouse {
     22 -> 0.045,
     23 -> 0.032,
   )
-  protected val waterVolumeRelativeFlat = Map(
+  private val waterVolumeRelativeFlat = Map(
     0 -> 0.01,
     1 -> 0.01,
     2 -> 0.01,
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -463,7 +463,7 @@ final case class ThermalHouse(
     )
 
     val durationValue = Math.log(
-      (nextInnerTemperatureToReach - longTermTemperature) / (currentInnerTemperature - longTermTemperature)
+      (nextInnerTemperatureToReach.toKelvinScale - longTermTemperature.toKelvinScale) / (currentInnerTemperature.toKelvinScale - longTermTemperature.toKelvinScale)
     ) / (k2 * -1)
 
     val duration = Math.floor(durationValue).toLong
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
@@ -12,6 +12,8 @@ import edu.ie3.datamodel.models.input.thermal.{
   ThermalBusInput,
   ThermalHouseInput,
 }
+import edu.ie3.util.scala.quantities.QuantityUtil.*
+import edu.ie3.simona.model.participant.hp.HpModel.HpState
 import edu.ie3.simona.model.participant.ParticipantModel.ModelState
 import edu.ie3.simona.model.thermal.ThermalGrid.ThermalEnergyDemand
 import edu.ie3.simona.model.thermal.ThermalHouse.ThermalHouseThreshold.{
@@ -28,10 +30,12 @@ import edu.ie3.util.scala.quantities.QuantityConversionUtils.toSquants
 import edu.ie3.util.scala.quantities.SquantsUtils.RichThermalCapacity
 import edu.ie3.util.scala.quantities.{ThermalConductance, WattsPerKelvin}
 import squants.energy.KilowattHours
-import squants.thermal.{Kelvin, ThermalCapacity}
+import squants.space.Litres
+import squants.thermal.{Celsius, Kelvin, ThermalCapacity}
 import squants.time.Seconds
-import squants.{Energy, Power, Temperature, Time}
+import squants.{Energy, Power, Temperature, Time, Volume}
 
+import java.time.ZonedDateTime
 import java.util.UUID
 
 /** A thermal house model
@@ -56,6 +60,10 @@ import java.util.UUID
   *   Lower temperature boundary
   * @param upperBoundaryTemperature
   *   Upper boundary temperature
+  * @param housingType
+  *   type of house, either `house` or `flat`
+  * @param houseInhabitants
+  *   number of people living in the building
   */
 final case class ThermalHouse(
     uuid: UUID,
@@ -68,6 +76,8 @@ final case class ThermalHouse(
     targetTemperature: Temperature,
     lowerBoundaryTemperature: Temperature,
     upperBoundaryTemperature: Temperature,
+    housingType: String,
+    houseInhabitants: Double,
 ) extends ThermalSink(
       uuid,
       id,
@@ -76,25 +86,26 @@ final case class ThermalHouse(
       bus,
     ) {
 
-  /** Calculates the energy demand at the instance in question by calculating
-    * the [[ThermalEnergyDemand]] to reach target temperature from current inner
-    * temperature. Since [[ThermalEnergyDemand]] is two parted, requiredEnergy
-    * and possibleEnergy, both have to be determined. RequiredEnergy: In case
-    * the inner temperature is at or below the lower boundary temperature, the
-    * energy demand till target temperature is interpreted as requiredEnergy.
-    * Otherwise, the required energy will be zero. PossibleEnergy: In case the
-    * inner temperature is not at or above the target temperature, the energy
-    * demand to reach targetTemperature is interpreted as possible energy.
-    * Otherwise, it will be zero. The current (external) thermal feed in is not
-    * accounted for, as we assume, that after determining the thermal demand, a
-    * change in external feed in will take place.
+  /** Calculates the energy demand for heating at the instance in question by
+    * calculating the [[ThermalEnergyDemand]] to reach target temperature from
+    * actual inner temperature. Since [[ThermalEnergyDemand]] is two parted,
+    * requiredEnergy and possibleEnergy, both have to be determined.
+    * RequiredEnergy: In case the inner temperature is at or below the lower
+    * boundary temperature, the energy demand till target temperature is
+    * interpreted as requiredEnergy. Otherwise, the required energy will be
+    * zero. PossibleEnergy: In case the inner temperature is not at or above the
+    * target temperature, the energy demand to reach targetTemperature is
+    * interpreted as possible energy. Otherwise, it will be zero. The current
+    * (external) thermal feed in is not accounted for, as we assume, that after
+    * determining the thermal demand, a change in external feed in will take
+    * place.
     *
     * @param thermalHouseState
     *   Current state, that is valid for this model.
     * @return
-    *   The needed energy in the questioned tick.
+    *   The needed energy for heating in the questioned tick.
     */
-  def energyDemand(
+  def energyDemandHeating(
       thermalHouseState: ThermalHouseState
   ): ThermalEnergyDemand = {
     // Since we updated the state before, we can directly take the innerTemperature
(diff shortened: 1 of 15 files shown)


src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
     ThermalEnergyDemand(requiredEnergy, possibleEnergy)
   }
 
-  /** Calculate the needed energy to change from start temperature to target
-    * temperature.
+  /** Checks if the simulation reached a new full hour in the simulation time.
+    * If we already simulated a tick within the current simulation time hour, it
+    * returns None. Else a sequence of the hours since the tick of state will be
+    * returned.
```
staudtMarius, line 131: Did I understand it correctly that we only check the demand once per hour?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+    * @param hour
+    *   The hour for that the energy demand should be determined.
+    * @return
+    *   The needed energy for domestic hot water in the questioned hour.
+    */
+
+  private def calculateThermalEnergyOfWaterDemand(hour: Int): Energy = {
+    val waterDemand =
+      waterDemandOfHour(hour, houseInhabitants, housingType)
+    thermalEnergyDemandWater(waterDemand, Celsius(10d), Celsius(55d))
```
staudtMarius: Why did you hardcode these values? Are that constants? If not maybe we should make them configurable via a `HotWaterStorageConfig` and set these values as defaults.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+    * @param waterDemand
+    *   water volume to get heated up
     * @param startTemperature
-    *   The starting temperature.
+    *   starting Temperature
+    * @param endTemperature
+    *   end Temperature
+    * @return
+    *   the needed energy
+    */
```
staudtMarius, line 218: This scaladoc need to be adapted to the new style.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+      endTemperature: Temperature,
+  ): Energy = {
+    if endTemperature < startTemperature then
+      throw new RuntimeException(
+        s"End temperature of $endTemperature is lower than the start temperature $startTemperature for the water heating system."
+      )
+
+    waterDemand.toCubicMeters * KilowattHours(
+      1.16
+    ) * (endTemperature.toCelsiusDegrees - startTemperature.toCelsiusDegrees)
```
staudtMarius: Could you add some comment, where this equation is coming from?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+    *
+    * @param hour
+    *   The hour for which the demand is to be calculated.
+    * @param noPersonsInHoushold
+    *   number of persons living in the building
+    * @param housingType
+    *   type of the building, either `house` or `flat`
+    * @return
+    *   the needed energy
+    */
```
staudtMarius: This scaladoc need to be adapted to the new style.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+
+    val isHouse: Boolean =
+      housingType.toLowerCase match {
+        case "house" => true
+        case "flat"  => false
+        case _ =>
+          throw new IllegalArgumentException(
+            s"Invalid housing type: $housingType. Expected 'house' or 'flat'."
+          )
+      }
```
staudtMarius, line 265: Maybe we should introduce an `enum` for this.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +124,231 @@ final case class ThermalHouse(
...
+    */
+
+  private def waterDemandOfHour(
+      hour: Int,
+      noPersonsInHoushold: Double,
+      housingType: String,
+  ): Volume = {
+
+    // Volume => VDI 2067 Blatt 12
+    // Time series relative => DIN EN 12831-3 Table B.2 for single family houses and for flats
```
staudtMarius: Does it make sense to move these constant maps to the companion object?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -235,12 +454,16 @@ final case class ThermalHouse(
...
+    * @param nextTryForPossibleThreshold
+    *   Next or initial possible threshold value that should be find by
+    *   recursive search.
     * @return
     *   The next threshold, that will be reached.
     */
   def determineNextThreshold(
       thermalHouseState: ThermalHouseState,
       qDot: Power,
+      nextTryForPossibleThreshold: Long = 1,
```
staudtMarius: This is currently unused. I guess this will be used in another PR.

src/test/scala/edu/ie3/simona/model/participant/hp/HpModelSpec.scala
```diff
@@ -174,6 +175,7 @@ class HpModelSpec
           updatedState match {
             case HpState(
                   tick,
+                  _,
```
staudtMarius, line 178: Does it make sense to check the time of the state here?

staudtMarius (changes requested):
Looks already quiete good. I just have some small remarks and quesstions.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: Yes. There is only hourly data available so far.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: These are based on VDI 2067 Blatt 12 - I don't see the necessity to make these configurable. On stochastic basis temperature preferece of people will be in that range.
I will add the reference in the code.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: Not sure what the new style is, please give me a hint if my adaptions doesn't match.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: This is the basic energy demand to heat 1m³ of water by one celsius. I don't want to add basics like that in the code. But this should be done in squants as well. So maybe then it becomes clearer.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: What would be the advantage?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: In my opinion this should start in psdm then. And maybe it's not worth the effort for just this two, or? What do you think.

src/test/scala/edu/ie3/simona/model/participant/hp/HpModelSpec.scala
danielfeismann: We're checking for the tick. That should be sufficient in my point of view. If would just increase the amount of changes one has to do when we need to change something here.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: No, no longer. Good catch!

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: Ok. I think, we can keep it this way and maybe adapt it, if we use other data in the future.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: Maybe you could put them as static constants in the companion object.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: An upper case letter at the begining and a `.` at the end.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: Do you expect to have more types in the future? I agree that this should be handled by the PSDM.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: I think defining constant values outside a method makes more sense.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: I looked this up and I think this is a bit misleading. Maybe we should add a comment that we precalcuated this specific ernergy demand. Normaly, one would use the actual specific heat demand with the unit J/gK.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: No, both values are from the standard. So in case we don't have another source, not.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: Well it's the same (4184 J/kg = 1.16 Wh/kg). We have the 1.16 also within the inputdata of cylindricalStorage. So imho now working with joules would be more confusing.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: I didn't get, that this is now fixed as new style. But ok :)

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
staudtMarius: The input unit in the PSDM is `kWh/(K * m³)`. Since we are using water and assume a constant density, we can use `kWh/(K * kg)`.

Since the unit of temperatur is Kelvin. We should adapt this in the code. This will not change the result, but we schould use the correct units.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,23 +137,177 @@ final case class ThermalHouse(
...
+      waterDemand: Volume,
+      startTemperature: Temperature,
+      endTemperature: Temperature,
+  ): Energy = {
+    if endTemperature < startTemperature then
+      throw new RuntimeException(
+        s"End temperature of $endTemperature is lower than the start temperature $startTemperature for the water heating system."
+      )
+
+    val specificHeatDemandWater = KilowattHoursPerCubicMeter(1.16)
```
staudtMarius: This is not the correct unit. Could you change this into `SpecificHeatCapacity`?

src/main/scala/edu/ie3/util/scala/quantities/SquantsUtils.scala
```diff
@@ -35,6 +35,14 @@ object SquantsUtils {
...
   }
 
+  implicit class RichEnergyDensity(energyDensity: EnergyDensity) {
+    def calcEnergyToHeat(
+        volume: Volume,
+        temperatureDelta: Temperature,
+    ): Energy = KilowattHours(
+      this.energyDensity.toKilowattHoursPerCubicMeter * volume.toCubicMeters * temperatureDelta.toCelsiusDegrees
+    )
+  }
```
staudtMarius: Can be removed, if we switch to the correct unit in `ThermalHouse`.

src/test/scala/edu/ie3/util/quantities/SquantsUtilsSpec.scala
```diff
@@ -34,6 +37,24 @@ class SquantsUtilsSpec extends AnyFlatSpec with Matchers {
...
+  }
+
+  it should "return Energy when an EnergyDensity is multiplied by Volume and temperature delta in Kelvin" in {
+    KilowattHoursPerCubicMeter(1.16).calcEnergyToHeat(
+      CubicMeters(4),
+      Kelvin(12),
+    ) should approximate(
+      KilowattHours(55.68)
+    )
+  }
```
staudtMarius: This could also be removed, if we change the unit in `ThermalHouse`.

staudtMarius (changes requested):
I have some requests regarding the unit that is used for `SpecificHeatCapacity`.

src/main/scala/edu/ie3/util/scala/quantities/SquantsUtils.scala
```diff
@@ -32,9 +32,18 @@ object SquantsUtils {
...
+  implicit class RichSpecificHeatCapacity(
+      specificHeatCapacity: SpecificHeatCapacity
+  ) {
+    def calcEnergyToHeat(
+        volume: Volume,
+        temperatureDelta: Temperature,
+    ): Energy = KilowattHours(
+      specificHeatCapacity.toKilowattHoursPerKelvinCubicMeters * volume.toCubicMeters * temperatureDelta.toKelvinDegrees
+    )
+  }
```
staudtMarius: There is already a similar method in `SpecificHeatCapacity`. Maybe you could just use that method instead of introducing a new one.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,24 +136,176 @@ final case class ThermalHouse(
...
+      throw new RuntimeException(
+        s"End temperature of $endTemperature is lower than the start temperature $startTemperature for the water heating system."
+      )
+
+    val specificHeatDemandWater = KilowattHoursPerKelvinCubicMeters(
+      Joules(4184e3).toKilowattHours
+    )
+    val temperatureDelta = endTemperature - startTemperature
+
+    specificHeatDemandWater.calcEnergyToHeat(waterDemand, temperatureDelta)
```
staudtMarius: There is already a method (`calcEnergy`) that takes a volume and two temperatures. This method also takes care of calculating the temperature difference.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,24 +136,176 @@ final case class ThermalHouse(
...
+  }
+
+  /** Provides water demand of a building (house or flat) for a given hour of
+    * the day based on the housingType and the number of inhabitants
```
staudtMarius: The period at the end of the sentence is missing.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -113,24 +136,176 @@ final case class ThermalHouse(
...
+      housingType.toLowerCase match {
+        case "house" => true
+        case "flat"  => false
+        case _ =>
+          throw new IllegalArgumentException(
+            s"Invalid housing type: $housingType. Expected 'house' or 'flat'."
+          )
+      }
+
+    val waterVolumeRelative: Map[Int, Double] =
```
staudtMarius: Do you need this type?

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
```diff
@@ -313,6 +488,71 @@ final case class ThermalHouse(
 object ThermalHouse {
   protected def temperatureTolerance: Temperature = Kelvin(0.01d)
 
+  /** Temperature values are set constant and based on VDI 2067 Blatt 12.
+    */
+
+  protected val lowerTemperatureTapWater: Temperature = Celsius(10d)
+  protected val upperTemperatureTapWater: Temperature = Celsius(55d)
```
staudtMarius: The empty line can be removed.

staudtMarius (changes requested):
Some small requests.

src/main/scala/edu/ie3/simona/model/thermal/ThermalHouse.scala
danielfeismann: good point...

staudtMarius (dismissed):
LGTM.

staudtMarius (approved):
LGTM. Thanks for all the work.

Issue #1422, opened 2025-06-11
Remove message retrieval timeout from powerflow calculation


sebastian-peter:
Message timeouts (`ask`) have been increased from 3s to 30s, but on (memory-)limited machines this might still not be enough.
My suggestion would be to remove the timeout alltogether (which might mean we have to move away from the ask pattern). We very rarely rely on the ask pattern anyways and normally just react to received messages.

```
14:27:46.517 INFO  o.a.pekko.pattern.PromiseActorRef - Message [edu.ie3.simona.agent.grid.GridAgentMessages$GridPowerResponse] to Actor[pekko://Simona/temp/2$e] was not delivered. [1] dead letters encountered. If this is not an expected behavior then Actor[pekko://Simona/temp/2$e] may have terminated unexpectedly. This logging can be turned off or adjusted with configuration settings 'pekko.log-dead-letters' and 'pekko.log-dead-letters-during-shutdown'.
[...]

PR #1569, opened 2025-10-18
pekko message update #1422

kkarthik11:
pekko message retrieval update #1422


sebastian-peter:
Currently not easily feasible, as expressed in #1422

PR #1580, opened 2025-10-29
Abstracting `MathFlexOptions` to `EnergyBoundariesFlexOptions`, integrating optimization into `OptimizedFlexStrat`

### LINKED ISSUE #1572: Implement energy limit flex options and adapt optimization

sebastian-peter:
This PR introduces `EnergyBoundariesFlexOptions`, which separates the flex options from the optimization logic.

Also fixes an erroneous soft constraint penalty calculation.

Resolves #1572


src/main/scala/edu/ie3/simona/model/em/opt/PowerObjectiveFactory.scala
```diff
@@ -0,0 +1,112 @@
...
+
+    override def build(
+        totalPower: Expression,
+        target: Power,
+    )(using model: MPModel): Expression = {
+      val difference = totalPower - target.toKilowatts
+
+      val differenceAbs = absoluteValue(difference, "differenceAbs")
+
+      val segmentSize = lastSegment / segmentCount
```
danielfeismann: I don't get this, can you help me here

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
```diff
@@ -0,0 +1,179 @@
...
+  )
+
+  object AssetEnergyBoundaries {
+
+    /** Creating energy boundaries for a fixed power time series. Assumes
+      * equidistant power series entries - otherwise, results are not defined!
+      *
+      * @param powerSeries
+      *   The power time series (at equidistant ticks). Has to consist of at
+      *   least two entries.
```
danielfeismann, line 77: if the equidistance is important maybe this should be highlighted within the naming?

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,445 @@
...
+
+import java.util.UUID
+
+/** Energy management strategy that optimizes flexibility usage over a fixed
+  * amount of time steps into the future. Takes [[EnergyBoundariesFlexOptions]]
+  * as inputs, which provide the required energy boundaries and power limits for
+  * constraining asset operation.
+  *
+  * @param sampleTime
+  *   The amount of time between the steps.
```
danielfeismann, line 34: Would this be the unit to which you convert discrete time events? It might be helpful to provide some additional explanatory comments on this in general here or in the method header above.

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,445 @@
...
+  * @param predictionHorizon
+  *   The amount of time that is predicted into the future, i.e. the last step
+  *   is this amount of time away from the current point in simulation time.
+  *   Should be a multiple of [[sampleTime]].
```
danielfeismann, line 38: Why should it be a multiple. It looks like it would work also with different values

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,445 @@
...
+      val formerEnergy = formerEnergyLimits.map(_.getUpper).getOrElse(zeroKWh)
+      val currentEnergy = energyLimits.getUpper
+
+      val fixedPower = (currentEnergy - formerEnergy) / sampleTime
+      StepResults(Const(fixedPower.toKilowatts), None, None)
+    } else {
+      // we do have some flexibility at this point in time, model it
+
+      // we use charging efficiency for both charging and discharging,
+      // since we use the adapted storage model here
```
danielfeismann, line 227: Perhaps you could find a better description for 'the adapted storage model'. I'm not sure if we'll be able to match this naming in the future.

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,445 @@
...
+
+          model.add(newState := previousState + p * sampleTime.toHours)
+          None
+        } else {
+          // there are charging/discharging losses, thus use the full model
+
+          val pAbsMax =
+            energyBoundaries.powerLimits.getUpper.max(
+              -energyBoundaries.powerLimits.getLower
+            )
```
danielfeismann, line 273: Asuming there are different power limits of the unit (e.g. +5 kW and -3 kW, this could result that the storage later discharges by e.g. -4 kW, right? That did not sound right to me

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
```diff
@@ -0,0 +1,445 @@
...
+          Some(AbsValueSoftConstraint(p, pAbs, eta, sampleTime))
+        }
+
+      StepResults(p, Some(newState), softConstraint)
+    }
+
+  }
+
+  /** Creates flex options that are equivalent to the original with regard to
+    * optimization, which are able to be optimized with a linear model though.
```
danielfeismann:
```suggestion
    * optimization, but these can be optimized with a linear model.
```

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+                pMax = pMax,
+                etaCharge = Each(etaCharging),
+                etaDischarge = Each(etaDischarging),
+                0L,
+              )
+            )
+            val adaptedLimits = adaptedFlexOptions.energyLimits(0)
+
+            val adapted = AdaptedModel(
+              currentEnergy = -adaptedLimits.getLower,
```
danielfeismann, line 76: Is the minus right here?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+
+        {
+          objectiveContainer.softConstraints.foreach { constraint =>
+            withClue(constraint.getWarningMessage) {
+              constraint.getError should be < constraintTolerance
+            }
+          }
+
+          // discharging 5 kWh plus 1.25 kWh losses
+          batRes(0).pVal should approximate(-5)
```
danielfeismann: The switched sign is a bit confusing, maybe it is right because it refers to a change in point of view. However perhaps you see a change to adapt this? (Compared to the Seq you provided to create the AssetEnergyBoundaries)

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+          batRes(0).energyVal should approximate(-6.25)
```
danielfeismann: Negative energyVal sounds also confusing. I assume this is the amount taken from the energy of last step, right?

danielfeismann: or, when i check the next steps its more from the first one?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+
+          // charging 10 kWh minus 2 kWh losses
+          batRes(1).pVal should approximate(10)
+          batRes(1).energyVal should approximate(1.75)
```
danielfeismann: ok, not sure what energyVal is about. Is it the total amount taken and provided to the storage?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: ok, looks like this is the total amount taken and provided to the storage, right? A comment explaning this would be great :)

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+              constraint.getError should be < constraintTolerance
+            }
+          }
+
+          // possibly charging
+          batRes(0).pVal should be >= 0d
+          batRes(0).energyVal should (be >= 0d and be <= 12d)
+
+          // possibly charging, now we should have reached 24 kWh
+          // (12 kWh above starting energy)
```
danielfeismann: Is that right? Assuming time step of one hour with max 10 kW (minus losses) does not reach 24 kWh...

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+          batRes(1).pVal should be >= 0d
+          batRes(1).energyVal should approximate(12d)
```
danielfeismann: approx 8?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+
+          // we should've charged 12 kWh plus 3 kWh losses
+          batRes(0).pVal + batRes(1).pVal should approximate(15d)
```
danielfeismann: I assumed charging eta of 20% like above, but they changed since the the power seq effect them, right? Maybe you can add a comment with the current eta at each test?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+      }
+
+    }
+
+    "provided with battery flex options without losses" should {
+
+      given ticks: Seq[Long] =
+        Range.Long.inclusive(0, 4 * sampleTicks, sampleTicks)
+
+      // low efficiency for simplicity of the test
```
danielfeismann:
```suggestion
      // high efficiency for simplicity of the test
```

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,929 @@
...
+
+          batRes should have size 12
+
+          batRes.slice(0, 4).foreach {
+            _.energyVal should (be >= 0d and be < 20d)
+          }
+
+          batRes.slice(4, 6).foreach {
+            _.energyVal should approximate(20d)
+          }
```
danielfeismann, line 769: i do not get this one

danielfeismann (changes requested):
Some comments from my side, I'm happy to discuss them

src/main/scala/edu/ie3/simona/model/em/opt/PowerObjectiveFactory.scala
sebastian-peter: It's the size of one segment. We're given the number of segments and the value of the last segment. Thus we can determine how "wide" each segment is.

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
sebastian-peter: I'd rather not make value names too long. But I mention this in scaladocs for this reason.

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
sebastian-peter: I don't know what you mean with "unit"... It's just the amount of time between time steps. Time steps are defined in line 28-29 above. But maybe I can make the description more clear?

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
sebastian-peter: While the method does not throw any error, and there might still be optimization happening, I think it goes against the logic of the optimization to provide prediction horizons that are not multiples of the sample time. I fixed the tick range generation though, as it was not inclusive before.

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
sebastian-peter: I could've called it "linearized storage model", but since I don't have a monopoly on linearizing storage models, I thought this would not help much either. I also thought it might be clear what adaptation is meant here, given that I explain this in the scaladoc of `adaptEnergyBoundaries`.

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
sebastian-peter: No, this is just the upper boundary for the `pAbs` variable. It can help the optimization to give bounds on possible values, but of course the bounds cannot be too narrow. And since this is the absolute value, we also have to consider the lower power limit.

The example you give is still working because the `p` variable, defined above, is limited exactly like you would expect.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Yes. Explanation is the same as in EnergyBoundariesFlexOptions lines 166ff

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Discharging the battery represents a feed-in and is thus negative... I don't see how this differs in SIMONA in general?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Yes. Energy at current tick is 0, and changes from the accumulate over time. I think a description of this belongs to `AssetEnergyBoundaries`, where I did describe it. But maybe I can copy the explanation to the StepResults as well.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: The battery starts half full, hence my reference to the starting energy

sebastian-peter: Also, this is the result of the second time step already, so you have 2x 10 kW of excess power already

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: As I mention above, this is the result of the second step already

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Eta is 80%. But the comment is misleading, since we're checking power here, not energy. I'll adapt it.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: I think 0.8 is not particularly efficient for a battery storage, is it?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Maybe you can formulate a question then? :D This checks whether steps 4 and 5 result in a fully charged battery, which makes sense to me if you look at the pv and load time series (-12/0 at step 4, 0/0 at step 5).

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: But it is 1.0. Isn't it? :)

src/main/scala/edu/ie3/simona/model/em/opt/PowerObjectiveFactory.scala
danielfeismann: I think during my review this got clearer. Here I was confused because I did not get that this will be the value of the lastSegement, not e.g. its number. I would prefer lastSegementVal or similar. But I also see advantages in short variable names...

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
danielfeismann: Unit was the wrong term... interval would it have been, sorry. 

So, just for my understanding: all discrete time events we have will be sampled to this interval, right? Perhaps we can add this in the method header (line 31) or line 34? E.g. `This requires to convert the time series to fit into the set sampleTime.` or similar?

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
danielfeismann: See your point, in both cases for later persons this is hard to find the reference :D 

I don't have any better advise, so let's keep it as it is.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: Alright, yes. It is result(0) not result of timestep 0...

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: Correct eta 80 %, losses 20 %...

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
```diff
@@ -0,0 +1,935 @@
...
+            withClue(constraint.getWarningMessage) {
+              constraint.getError should be < constraintTolerance
+            }
+          }
+
+          // possibly charging
+          batRes(0).pVal should be >= 0d
+          batRes(0).energyVal should (be >= 0d and be <= 6d)
+
+          // possibly charging, now we should have reached 12 kWh
```
danielfeismann, line 362: If we still start at 12, we should have reached after two times 0.5h x power something above 12, right?

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: Sorry :D You're right and I missed one entry within loadFlex which confused me why the storage isn't used to cover the load...

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: That would be a help for future times working on this part of the test :)

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: I was refering to the Seq() you provide in line 149. The signs here are as I would expect them. Should have placed my comment above...

src/main/scala/edu/ie3/simona/model/em/opt/PowerObjectiveFactory.scala
sebastian-peter: We can call it `lastSegmentValue`, no problem

sebastian-peter: Or even better, `maximumExpectedPower` and make it a Power

src/main/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStrat.scala
sebastian-peter: The energy boundaries do not have to be sampled in this exact resolution, no. Although it's more efficient if that happens, of course.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Ok, well, what I call `constFlex` above is a fixed feed-in or load connected to the EM. So, if that source is consuming 5 kW (positive), the battery needs to feed-in 5 kW, which then becomes -5kW due to passive sign convention (Verbraucherzählpfeilsystem).

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Yes, this is changing with this PR - we don't explicitly return the initial state (0 kWh) anymore.

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Ok, well, it helps to look into the specific code. You're right of course, this is just a copy-paste error

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Sorry, I changed things again after your review, now we're starting at 6 kWh. Full capacity is 12 kWh, so it's still half energy. I did this because when only testing with sample time of one hour, one can miss some bugs... 🐞

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: Alright, perhaps it is possible to have the sequence as extra variable named fixedFeedIn, or some comment that this mock a fixed feed-in...

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
danielfeismann: Maybe it helps to link both defintions or add a short comment here as well. Because this is not directly intuitive to me

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Added comments

src/test/scala/edu/ie3/simona/model/em/opt/OptimizedFlexStratSpec.scala
sebastian-peter: Ok, I guess you'd like to have more commentary in the test, not only in StepResults. Added some to  `OptimizingTestLike`.

danielfeismann: Sorry, I planed to discuss this in office with you - but we always missed each other :D 
Hope we will meet the next days, will be a quick discussion I guess.

danielfeismann (approved):
LGTM Thanks for answering my questions and adding more information

PR #1637, opened 2026-01-15
Creating `DataInputHandler` for both participant and EM agent

### LINKED ISSUE #1631: Create abstract `DataInputHandler` from `ParticipantInputHandler`
Functionality is now also needed for `EmAgent`...

sebastian-peter:
Closes #1631

### DIFF
--- a/src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
+++ b/src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
@@ -7,6 +7,7 @@
 package edu.ie3.simona.agent.participant
 
 import breeze.numerics.{pow, sqrt}
+import edu.ie3.simona.agent.DataInputHandler
 import edu.ie3.simona.agent.grid.GridAgentMessages.{
   AssetPowerChangedMessage,
   AssetPowerUnchangedMessage,
@@ -84,7 +85,7 @@ object ParticipantAgent {
 
   /** Data provision messages sent by data services.
     */
-  sealed trait DataInputMessage extends Request {
+  sealed trait DataMessage extends Request {
 
     /** The current tick.
       */
@@ -110,7 +111,7 @@ object ParticipantAgent {
       override val serviceRef: ActorRef[ServiceMessage],
       data: Data,
       override val nextDataTick: Option[Long],
-  ) extends DataInputMessage
+  ) extends DataMessage
 
   /** Providing the information that no data will be provided by the sending
     * service for the current tick. The participant could thus potentially skip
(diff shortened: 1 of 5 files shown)


src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
```diff
@@ -307,38 +307,24 @@ object ParticipantAgent {
...
-      ParticipantModelShell[?, ?],
-      ParticipantInputHandler,
-      ParticipantGridAdapter,
-  ) = {
-    if expectedMessagesReceived(inputHandler, gridAdapter) then {
-
-      val activation = inputHandler.activation.getOrElse(
+  ): ParticipantAgentData = {
+    if expectedMessagesReceived(data) then {
+
```
staudtMarius, line 326: Does it makes sense to get and save the needed objects from the data. This would reduce the number of changed lines. It could also be a bit faster, since no inlining needs to be performed by  the JIT  compiler.

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
```diff
@@ -482,20 +479,17 @@ object ParticipantAgent {
...
-      inputHandler: ParticipantInputHandler,
-      gridAdapter: ParticipantGridAdapter,
+      data: ParticipantAgentData
   ): Boolean = {
-    inputHandler.allMessagesReceived &&
-    inputHandler.activation.exists(activation =>
-      !gridAdapter.isPowerRequestAwaited(activation.tick)
+    data.activation.exists(activation =>
+      !data.gridAdapter.isPowerRequestAwaited(activation.tick) &&
+        data.inputHandler.allMessagesReceived(activation.tick)
```
staudtMarius, line 492: Why did you change the order here?

staudtMarius (changes requested):
Just two small comments from my side.

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
sebastian-peter: Because you need to hand over the activation tick to `allMessagesReceived` now

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
sebastian-peter: I had that before, but I like it more that the way it is now, we can better distinguish between the input data and updated input data. Furthermore, I think we shouldn't concern ourselves with manual JIT optimizations at all.

staudtMarius (approved):

PR #1638, opened 2026-01-15
Added filter function for unchanged grid results

### LINKED ISSUE #487: Only persist switch result when switch position changes

staudtMarius:
Resolves #487.

### DIFF
--- a/src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
+++ b/src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
@@ -57,6 +57,10 @@ object ResultServiceProxy {
     */
   final case class AddListener(listener: ActorRef[ResultResponse])
 
+  given Conversion[Map[UUID, ResultEntity], Map[UUID, Iterable[ResultEntity]]] =
+    (inputMap: Map[UUID, ResultEntity]) =>
+      inputMap.map { case (key, value) => key -> Iterable(value) }
+
   /** State data of the [[ResultServiceProxy]].
     * @param listeners
     *   A sequence of listeners. The proxy will forward all results to these
@@ -91,7 +95,7 @@ object ResultServiceProxy {
         Transformer3wKey,
         AggregatedTransformer3wResult,
       ] = Map.empty,
-      gridResults: Map[UUID, Iterable[ResultEntity]] = Map.empty,
+      gridResults: Map[UUID, ResultEntity] = Map.empty,
       results: Map[UUID, Iterable[ResultEntity]] = Map.empty,
       waitingForResults: Map[UUID, Long] = Map.empty,
       requiresSetPoint: Set[UUID] = Set.empty,
@@ -218,7 +222,7 @@ object ResultServiceProxy {
       uuids.flatMap { uuid =>
         gridResults.get(uuid) match {
           case Some(values) =>
-            Some(uuid -> values)
+            Some(uuid -> Iterable(values))
           case None =>
             results.get(uuid).map { res => uuid -> res }
         }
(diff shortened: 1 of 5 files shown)


src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
danielfeismann:
```suggestion
  /** Method for adding a listener to the proxy.
```

src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
```diff
@@ -57,6 +57,10 @@ object ResultServiceProxy {
     */
   final case class AddListener(listener: ActorRef[ResultResponse])
 
+  given Conversion[Map[UUID, ResultEntity], Map[UUID, Iterable[ResultEntity]]] =
```
danielfeismann, line 63: Could you add the implicit conversion also to the scala doc above?

src/test/scala/edu/ie3/simona/service/results/ResultServiceProxySpec.scala
```diff
@@ -199,15 +211,47 @@ class ResultServiceProxySpec
...
+
+      resultProxy ! PowerFlowResultEvent(
+        Seq(dummyNodeResult),
+        Seq(dummySwitchResult),
+        Seq(dummyLineResult),
+        Seq(dummyTrafo2wResult),
+        Seq.empty,
+      )
+
+      // all results have the same uuid, therefore, all result a grouped to this uuid
```
danielfeismann: do they have the same uuid?

src/test/scala/edu/ie3/simona/service/results/ResultServiceProxySpec.scala
```diff
@@ -55,10 +55,16 @@ class ResultServiceProxySpec
...
 
       resultProxy ! RequestResult(
-        Seq(dummyInputModel, inputModel),
+        Seq(
+          dummyNodeResultModel,
+          dummySwitchResultModel,
+          dummyLineResultModel,
+          dummyTrafo2WResultModel,
+          inputModel,
+        ),
```
danielfeismann: this seq can be stored in a variable above and some lines may be saved.

src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
```diff
@@ -337,9 +341,30 @@ object ResultServiceProxy {
...
       val gridResults =
-        (transformer3wResults ++ nodeResults ++ switchResults ++ lineResults ++ transformer2wResults ++ congestionResults)
-          .groupBy(_.getInputModel)
+        (transformer3wResults ++ nodeResults ++ switchResults ++ lineResults ++ transformer2wResults ++ congestionResults).flatMap {
+          res =>
+            val model = res.getInputModel
+
+            oldResults.get(model) match {
+              case Some(oldResult) =>
+                // save the old time
```
danielfeismann: to me this would be clearer: "set old time to current result to allow comparison"

danielfeismann: or 
```
// Temporarily change time for comparison, then revert
      val oldTime = oldResult.getTime
      oldResult.setTime(res.getTime)
      val equal = oldResult == res
      oldResult.setTime(oldTime)

      if (equal) None else Some(model -> res)
```

danielfeismann (changes requested):
Thanks for tackling this issue. Some minor improvements from my point of view.

src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
staudtMarius: Which scaladoc do you mean?

src/test/scala/edu/ie3/simona/service/results/ResultServiceProxySpec.scala
staudtMarius: No, but previously they were.

danielfeismann: So does this comment need some update?

src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
danielfeismann: the lines above, line 56/57. 

If I remeber right, we nowadays also provide some information on the implicits / given there.

src/test/scala/edu/ie3/simona/service/results/ResultServiceProxySpec.scala
danielfeismann: Nevermind, did not see the changes you already did.

src/main/scala/edu/ie3/simona/service/results/ResultServiceProxy.scala
staudtMarius: @danielfeismann I added some scaladc to the conversion itself. Should I add it at another spot too?

danielfeismann: No, it is fine now, thanks!

danielfeismann (approved):
LGTM

PR #1655, opened 2026-01-22
Harmonizing test resource retrieval

sebastian-peter:
Closes #1654

PR #1518, opened 2025-09-04
Enhancing `WeatherService` to provide forecasts

sebastian-peter:
Resolves #1512 

Merge first:
- #1517
- #1525

PR #1656, opened 2026-01-22
Implementing price service

### LINKED ISSUE #1586: Implement `EnergyPriceService`
This should also pre-calculate forecasts and prosumer-specific buy and sell prices

sebastian-peter:
Closes #1586 

Merge first:
- #1655


sebastian-peter:
@staudtMarius you're very welcome to review whenever you have time
@danielfeismann if you have time, I'd be glad if you could look my additions in `config.md` and the price handling in `EnergyPriceService` and if that makes sense to you.

src/main/scala/edu/ie3/simona/config/InputConfig.scala
```diff
@@ -177,4 +184,46 @@ object InputConfig {
...
+  /** Class with parameters for price adjustments given a wholesale price.
+    *
+    * @param fees
+    *   The fees (in EUR/MWh) to be paid by the consumer or producer.
+    * @param tax
+    *   The tax to be paid by the consumer or producer, relative to the net
+    *   price (market price + fees).
+    */
+  final case class PriceAdjustments(
+      fees: Double = 0d,
```
staudtMarius, line 236: Could you add the default values to the scaladoc and the config doc (see: Default configuration values)?

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+    *   The definition of the source to use.
+    * @param startDateTime
+    *   Date and time of the very first tick in the simulation.
+    */
+  final case class InitPriceServiceStateData(
+      sourceDefinition: InputConfig.PriceDatasource,
+      startDateTime: ZonedDateTime,
+  ) extends InitializeServiceStateData
+
+  /** State data of an initialized weather service.
```
staudtMarius: ... price service.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+            val fileNamingStrategy = new FileNamingStrategy()
+
+            val metaSource = new CsvTimeSeriesMetaInformationSource(
+              csvSep,
+              Paths.get(directoryPath),
+              fileNamingStrategy,
+            )
+
+            val allMetaData =
+              metaSource.getTimeSeriesMetaInformation.values.asScala
```
staudtMarius: Instead of a meta source, you could directly use a `CsvDataSource`. Then you could use the method `getCsvIndividualTimeSeriesMetaInformation`.

With this you could only retrieve those meta data that have the given column scheme.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+                .filter(_.getColumnScheme == ColumnScheme.ENERGY_PRICE)
+
+            allMetaData
+              // taking the first one
+              .headOption
+              .map { metaData =>
+                if allMetaData.size > 1 then
```
staudtMarius: Does it makes sense to configure the price time series via the config? For example by specifing the uuid of the time series?

Combined with the use of `CsvDataSource` it could look like this: `Option(dataSource.getCsvIndividualTimeSeriesMetaInformation(ColumnScheme.PRICE).get(priceSeriesUuid))`

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+      case invalidMessage =>
+        Failure(
+          InvalidRegistrationRequestException(
+            "Cannot register an agent for weather service with registration " +
+              s"request message '$invalidMessage'!"
+          )
+        )
+    }
+
+  /** Try to register the sending agent with its coordinate and weather data
```
staudtMarius: The scaladoc needs to be adapted.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+            .to(SortedMap)
+
+          val priceSeries =
+            reduceTimeSeriesResolution(valueSeries, resolution).map {
+              case (tick, value) =>
+                tick -> createProsumerPrice(value)
+            }
+
+          SecondarySeriesData(
+            reduceTimeSeriesResolution(priceSeries, resolution)
```
staudtMarius: Why are you calling the method `reduceTimeSeriesResolution` again?

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+    )
+
+    val sellingPrice =
+      calculateSellingPrice(wholesalePrice, stateData.config.sellingPrice)
+    val buyingPrice =
+      calculateBuyingPrice(wholesalePrice, stateData.config.buyingPrice)
+    ProsumerPrice(sellingPrice, buyingPrice)
+  }
+
+  private given priceConv: Conversion[EnergyPriceValue, Option[EnergyPrice]]
```
staudtMarius: The conversion could be simplyfied to: `priceConv: Conversion[EnergyPriceValue, Option[EnergyPrice]] = (value: EnergyPriceValue) => value.getPrice.toScala.map(_.toSquants)`

But do you even use this conversion anywhere, or could you simply define this as a method?

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
```diff
@@ -0,0 +1,399 @@
...
+      adjustments: PriceAdjustments,
+  ): EnergyPrice =
+    (wholesalePrice + EuroPerMegawattHour(adjustments.fees)) *
+      (1d + adjustments.tax)
+
+  private def calculateSellingPrice(
+      wholesalePrice: EnergyPrice,
+      adjustments: PriceAdjustments,
+  ): EnergyPrice =
+    (wholesalePrice - EuroPerMegawattHour(adjustments.fees)) *
```
staudtMarius, line 375: Shouldn't this be `(wholesalePrice - EuroPerMegawattHour(adjustments.fees)) / (1d + adjustments.tax)`?

src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
```diff
@@ -283,11 +277,11 @@ object WeatherService extends SimonaService {
...
       coordinateData.registeredActors.foreach { case (dataType, actors) =>
         val weatherData = dataType match {
-          case WeatherDataType.Current =>
+          case SecondaryDataType.Current =>
             updatedStateData.weatherSource.getWeather(tick, coordinateWeights)
-          case WeatherDataType.CurrentAndForecast(length, interval) =>
+          case SecondaryDataType.CurrentAndForecast(length, resolution) =>
             val endTick = tick + length.toSeconds.toLong
-            // weather time series is forwarded without adding error
+            // price time series is forwarded as forecast without adding noise
```
staudtMarius: The change here is wrong.

src/test/scala/edu/ie3/simona/service/weather/WeatherServiceSpec.scala
```diff
@@ -168,17 +148,17 @@ class WeatherServiceSpec
...
         agent1.ref,
         WeatherRegistrationData(
           Coordinate(validCoordinate.latitude, validCoordinate.longitude),
-          WeatherDataType.Current,
+          SecondaryDataType.Current,
         ),
       )
 
       agent1.expectNoMessage()
+      agent2.expectNoMessage()
```
staudtMarius, line 161: Is this necessary?

src/test/scala/edu/ie3/simona/service/price/EnergyPriceServiceSpec.scala
```diff
@@ -0,0 +1,180 @@
...
+      )
+
+      agent1.expectNoMessage()
+    }
+
+    "send out correct weather information upon activity start trigger and request the triggering for the next tick" in {
+      /* Send out an activity start trigger as the scheduler */
+      priceService ! Activation(0)
+
+      val activationMsg = scheduler.expectMessageType[Completion]
```
staudtMarius: Could you change the name to `completionMsg`?

staudtMarius (changes requested):
Here are some comments from my side.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: Oh, embarrassing

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: Because we might want forecast with a coarser resolution than the source provides? It originates in #1518 for weather forecasts

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: I intended to use the conversion at first, now it's a method

src/main/scala/edu/ie3/simona/service/weather/WeatherService.scala
sebastian-peter: Good catch, I was sleeping there

src/test/scala/edu/ie3/simona/service/weather/WeatherServiceSpec.scala
sebastian-peter: I'd leave it in to catch weird behavior

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: Don't think so. If sales tax is 19% and your gross revenue is 1€, then you have to pay 19 cents tax and keep 81 cents net revenue.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: Originally I didn't want to implement this but this actually simplified and solidified things a bit

docs/readthedocs/config.md
```diff
@@ -74,13 +74,61 @@ simona.input.weather.datasource = {
...
   - The parameter `maxCoordinateDistance` is used to specify the radius in which weather data should be searched in. The given distance should be in meter.
 
 Further model classes which can be used to parse a data set as input to power system simulations are described in [PSDM](https://powersystemdatamodel.readthedocs.io/en/latest/models/models.html#time-series). 
 Data sources and data sinks are explained in the [I/O-capabilities](https://powersystemdatamodel.readthedocs.io/en/latest/io/basiciousage.html) section of the PSDM. 
 
-### Output parameters
+### Price data
+
+For the buying price, fees (in EUR/MWh) are added to the wholesale price and the tax is added on the total cost.
+Conversely, for the selling price, fees (in EUR/MWh) are subtracted from the wholesale price, and the tax is subtracted from the remaining amount.
```
danielfeismann, line 86: Current German regulation does not include grid fees for feed-in. However this might change since there is some discussion on [updating the current scheme](https://www.bundesnetzagentur.de/1059400). Other countries have this. Since SIMONA foremost but not only covers the German energy system, we might find some better solution on this? Maybe grid fees (absolute? relative?) can be provided by config separated for consumtion and feed-in?

docs/readthedocs/config.md
```diff
@@ -74,13 +74,61 @@ simona.input.weather.datasource = {
...
+
+Find below an exemplary configuration for households in Dortmund, 2025.
+The UUID of the price time series to use as data source. An individual time series with given UUID and column scheme ENERGY_PRICE needs to be provided.
+
+```
+simona.input.prices.datasource = {
+  buyingPrice = {
+    fees: 187.11
+    tax: 0.19
```
danielfeismann, line 97: Guess this is percent and represents the VAT? There also other parts, like Stromsteuer. However that one is absolute. 

I would avoid creating some mess here. So one option could be to set up this first for Germany and continue in another step by generalising this in a smart way?

docs/readthedocs/config.md
```diff
@@ -74,13 +74,61 @@ simona.input.weather.datasource = {
...
+  csvParams = {
+    directoryPath: "input/samples/vn_simona/fullGrid"
+    csvSep: ","
+    isHierarchic: false
+  }
+}
+```
+
+The exemplary price adjustments above can be explained as follows.
+Recurring fixed costs, such as metering point operation (Messstellenbetrieb) and the basic charge of grid fees (Grundpreis der Netzentgelte) are not considered.
```
danielfeismann: Understandable decission. However, enable this would provide the option to simulate for different locations correctly to show effects caused by higher or lower fees, etc..

danielfeismann: This will even get worse, since the basic charge of grid fees does not exist explictly. In my understanding this is the implicit result when turn over the capacity price (€ / MW) to LV customers since at least in Germany the capacity is not charged individually. So, for loads in MV and above this would be easy and I would prefer to have the prices in the structure of capacity price and commodity price. But this would lead to the next question how to deal with the separation based on the full load hours? Also this may can be done for some LV customers (e.g. commercial buildings etc..) when they are above 100 MWh/year since that would require constant power measurement (Registrierende Lastgangsmessung/Leistungsmessung). But, for classic households the fixed part will be the prefered option.

docs/readthedocs/config.md
```diff
@@ -74,13 +74,61 @@ simona.input.weather.datasource = {
...
+        - Offshore-Umlage: 0.816 ct/kWh
+        - NEV-Umlage/Aufschlag für besondere Netznutzung: 1.558 ct/kWh
+        - KWKG-Umlage: 0.277 ct/kWh
+        - Concession levy (Konzessionsabgabe): 2.39 ct/kWh
+        - Supplier margin (Lieferantenmarge): 4.45 ct/kWh
+    - Tax
+        - Sales tax (Umsatzsteuer): 19%
+- Selling price
+    - Fees
+        - Not standardized, thus we assume 0.5 ct/kWh
```
danielfeismann: There is also a second energy that is sold, in case it is under the subsidy of EEG (and there are also others like KWK). The EEG part is now covered by KTF which makes it equal zero for the buyer. But those DERs who are not selling their energy because they apply to subsidy, will get the respective (and individually different) price.

docs/readthedocs/config.md
```diff
@@ -74,13 +74,61 @@ simona.input.weather.datasource = {
...
+    - Tax
+        - Sales tax (Umsatzsteuer): 19%
```
danielfeismann:
```suggestion
        - Sales tax (VAT / Umsatzsteuer): 19%
```

danielfeismann: Since the most will not sell the energy to the endcustomer directly, this does not apply here in my opinion since the VAT will be reimbursed (Durchlaufender Posten...)

danielfeismann (dismissed):
The challenges here are even greater than I had imagined from your comments during our recent conversation. Only checked config.md. Please see my comments as my own notes when thinking about this. We can start with this as some first simple proposal, but should mention the limitations and challenges introduced by it then. However, it would be great to have some structured discussion on this topic the next weeks since I think with some effort a solution can be found that combines not too much detailed configuration while providing some system that can be used for different simulation cases...

docs/readthedocs/config.md
danielfeismann: I would prefer to keep tax completely out for both sides. Of course they have an impact on results but many effects may can be analysed better without been effected by taxation?

danielfeismann: I would prefer to keep tax completely out for both sides. Of course they have an impact on results but many effects may can be analysed better without been effected by taxation?

docs/readthedocs/config.md
sebastian-peter: Yes, this how my proposal intends to work. For both consumption and feed-in, one is supposed to set absolute and relative additions/subtractions. The absolute part I called "fees" for lack of a better umbrella term, the relative part (added/subtracted after fees are applied) I called "tax", which of course one can discuss as well. That's why, for example, for consumption, Stromsteuer is listed under fees, while Umsatzsteuer is a tax.

In any way, my idea was to sum up absolute and relative parts for both consumption and feed-in and list them as "fees" and "tax", even though the terms might not be perfect.

docs/readthedocs/config.md
sebastian-peter: See my comment above - Stromsteuer is included in "fees" here since it's absolute. We could rename them fees -> absolute and tax -> relative, however, I'm not convinced that this would be much less confusing.

And yes, I'd like to primarily set up an exemplary configuration for Germany here - to be exact, for Dortmund in 2025, since for example, grid fees depend on the grid operator and change from year to year. Other configurations can be easily derived then.

docs/readthedocs/config.md
sebastian-peter: I am aware that one can make the pricing model infinitely more comprehensive and sophisticated. Just your first suggestion here, differentiating between prices of different municipalities, would be very interesting. However, for now, I'd like to focus on small prosumer households and not complicate things too much. Enhancements are very much welcome after that.

docs/readthedocs/config.md
sebastian-peter: Do you mean the "EEG-Marktprämie"? In my understanding, only DERs that apply for EEG can opt for that if they directly market their energy. I admit that I have ignored this so far, as it seems to be highly individual, based on capacity, installation date etc. 
Maybe we can attempt to find some average value and subtract it from the "fee"? In general, I'm currently more interested to have a somewhat-believable differentiation between buying and seeling price, instead of covering individual cases.

docs/readthedocs/config.md
sebastian-peter: Ok, makes sense for the selling side (why this is "Durchlaufender Posten" is beyond me though - one is selling energy that one supplies from their own installation, why is it "durchlaufend"? Because the sun is the actual originator? We don't reimburse the sun though).
My difficulties with the concept aside, I'm willing to accept that we don't have VAT for the selling side. However, for the buying side, I thought it has to be paid anyway?

sebastian-peter:
> The challenges here are even greater than I had imagined from your comments during our recent conversation. Only checked config.md. Please see my comments as my own notes when thinking about this. We can start with this as some first simple proposal, but should mention the limitations and challenges introduced by it then. However, it would be great to have some structured discussion on this topic the next weeks since I think with some effort a solution can be found that combines not too much detailed configuration while providing some system that can be used for different simulation cases...

Thanks for taking your time to share your thoughts. We should definitely make sure to declare the limitations of this approach.

docs/readthedocs/config.md
danielfeismann: Alright, thanks for explaining a bit more. Maybe it was also to late to read this :-) How about just call the positions `absolute fees and taxes` and `relative fees and taxes`?

docs/readthedocs/config.md
danielfeismann: Agreed :) Just wanted to make sure that we don't forget this even I was sure you had this already on your radar.

docs/readthedocs/config.md
danielfeismann: I was refering to classic EEG, where EEG Marktprämie would be another thing which complicates this. However, I agree this is highly individual thus there should be some pragmatic solution as this is the overall target. I'm fine with skipping this at all at first. On the other hand, maybe some easy approach with just fixed renumeration (not price sensitive) would do it and help to cover this different market behaviour of these plants as well.

docs/readthedocs/config.md
danielfeismann: Let's disucss this the next days. Short: Imho, VAT in general only applies when some good is sold to an end customer. So I assume this applies here as well. But we should check these constellations here.

danielfeismann:
> > The challenges here are even greater than I had imagined from your comments during our recent conversation. Only checked config.md. Please see my comments as my own notes when thinking about this. We can start with this as some first simple proposal, but should mention the limitations and challenges introduced by it then. However, it would be great to have some structured discussion on this topic the next weeks since I think with some effort a solution can be found that combines not too much detailed configuration while providing some system that can be used for different simulation cases...
> 
> Thanks for taking your time to share your thoughts. We should definitely make sure to declare the limitations of this approach.

Yes. And maybe this should be done in some separate .md file and section of the docs. Since I don't see this as part of the docs for config.

src/main/scala/edu/ie3/simona/config/InputConfig.scala
```diff
@@ -32,6 +39,7 @@ final case class InputConfig(
     loadProfile: LoadProfile = LoadProfile(),
     primary: Primary = Primary(),
     weather: Weather = Weather(),
+    prices: Prices = Prices(),
```
staudtMarius, line 44: Could you add this to the scaladoc?

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
staudtMarius: I understand this point, but why are we calling this method two times with the same input parameters?

staudtMarius (commented):
Two more small remarks from my side.

src/main/scala/edu/ie3/simona/service/price/EnergyPriceService.scala
sebastian-peter: Ah... Good catch. That doesn't make sense of course

src/main/scala/edu/ie3/simona/agent/SecondaryServiceRegistration.scala
```diff
@@ -201,9 +201,10 @@ trait SecondaryServiceRegistration[Msg >: ServiceMessage.Response, CR] {
         }
 
       case ServiceType.PriceService =>
-        throw new CriticalFailureException(
-          s"${assetInput.identifier} is trying to register for a ${ServiceType.PriceService}, " +
-            s"which is currently not supported."
+        serviceRef ! SecondaryServiceRegistrationMessage(
+          registrantRef,
+          DataTimeType.Current,
+          (),
```
staudtMarius, line 207: I don't like the `()` here. Couldn't you just send the uuid of the asset as we do for the movement service?

Or we could make this parameter an option with a default value of `None`.

staudtMarius (commented):
One last comment from my side. If you think that its not important, you can ignore the comment,

@danielfeismann Could you have a look, if your points have been taken care of`?

src/main/scala/edu/ie3/simona/agent/SecondaryServiceRegistration.scala
sebastian-peter: [`()` is the value of type `Unit`](https://www.scala-lang.org/api/current/scala/Unit.html) and fits perfectly here, imho. The UUID wouldn't be used by the price service and an Option is also unnecessary I think.

sebastian-peter:
> @danielfeismann Could you have a look, if your points have been taken care of`?

I think we wanted to have some discussion before this gets merged...

danielfeismann (commented):
I'm fine with this for now :)

staudtMarius (approved):

PR #1674, opened 2026-02-03
Providing extra functionality for `EnergyBoundariesFlexOptions`

sebastian-peter:
Closes #1670

PR #1676, opened 2026-02-03
Adapting EM initialization protocol for requirements of energy boundaries

### LINKED ISSUE #1669: Adapt EM initialization protocol
Introduce `FlexInit` instead of `FlexActivation(INIT_SIM_TICK)`, among others.

sebastian-peter:
Closes #1669

Merge first:
- #1674

We could also think about renaming `FlexActivation` back to `FlexRequest`, since that's what it's doing now again. I don't know how much effort this would make overall though.

@staudtMarius Please also check whether the changes of this PR work for the EM services and simonaAPI connections that you implemented.


src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgentInit.scala
```diff
@@ -166,20 +177,36 @@ object ParticipantAgentInit
...
 
   }
 
   /** Waits for the primary proxy to respond, which decides whether this
     * participant uses model calculations or just replays primary data.
     */
-  private def waitingForPrimaryProxy(using
+  private def waitingForPrimaryProxy(
+      flexType: Option[FlexType],
+      dataTimeType: DataTimeType,
```
staudtMarius, line 215: Could you add these parameters to the scaladoc?

staudtMarius (changes requested):
LGTM. Just one small remark.

src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgentInit.scala
sebastian-peter: Sure

src/main/scala/edu/ie3/simona/model/participant/ParticipantModelShell.scala
```diff
@@ -527,10 +568,78 @@ object ParticipantModelShell {
   ): ParticipantModelShell[OP, S] =
     new ParticipantModelShell(
       model = model,
+      // required because FlexModelShellImpl cannot be created within constructor.
```
staudtMarius: What's the reason for this? Couldn't you move the impl class to the object?

staudtMarius (changes requested):
One more question from my side.

src/main/scala/edu/ie3/simona/model/participant/ParticipantModelShell.scala
sebastian-peter: Then you don't have access to the members of the `ParticipantModelShell` class, namely identifier and simulationStart. But honestly I never liked my solution here very much. Will try to move everything to a separate class, give me a couple of minutes.

staudtMarius (approved):
LGTM.

Issue #1096, opened 2025-01-09
Adapt interpretation of minimum SOC for EVCS


sebastian-peter:
Currently, we're forcing EVs to charge until they reached some minimum SOC, and above that level EMs are free to do whatever they want, which leads to large differences in charged energy compared to non-em-controlled behavior.

What we want instead is something like this:
EVCS is free to do what it wants with EVs (charge, discharge if v2g is enabled, do nothing). At some point in time though, we force charging _before_ the EV leaves (we know the departure time). This point in time is exactly the last moment where charging with full power leads to an EV with minimum SOC.

To have a somewhat energy-equivalent behavior when not em-controlling the EVCS, maybe the regular (non-em-controlled) behavior would be to charge until minimum SOC as well, and then stop charging. 

Discovered by @danielfeismann.

danielfeismann:
I couldn't have described it better. Thank you for creating this issue @sebastian-peter. In order to be flexible for further developments, maybe we should give a target SoC that the car could give optionally, otherwise the target SoC would be 100% I guess, right?

sebastian-peter:
@danielfeismann Yes, sounds right.

And just to spell it out: For discharging with v2g, the target SOC should not be relevant, right? So there should not be a necessity to discharge an EV with 100% SOC because the target is 80%?

danielfeismann:
No, I don't think so. But we may should think about some minimum level that never should be touched in case v2g takes place?

But we should be careful not over engineering this...

sebastian-peter:
In that case we'd have the minSOC as it is now, and additionally a targetSOC. Makes logical sense, I'll think about if this would be too cumbersome to implement

danielfeismann:
I would like to adapt this as well so we have the possibility either to provide a global upper and lower limit as well as get individual ones for each vehicle. So it would be possible that the MobilitySim provide e.g. the distance of the next trip which would allow to estimate a minimum SoC. But imho this is not a must feature...

sebastian-peter:
Addressing the target charging functionality in #1696 (basically what I described in the first post of this issue). For further developments, new issues would need to be created.

cc @danielfeismann

PR #1689, opened 2026-02-10
Implementing energy boundaries flex options for EVCS

### LINKED ISSUE #1682: Implement energy boundaries flex options for EVCS

sebastian-peter:
Closes #1682 

Merge first:
- #1676

This also touches upon the idea behind #1096, just not for PowerLimitFlexOptions but EnergyBoundariesFlexOptions. Edit: #1096 has now been implemented and merged


src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
```diff
@@ -0,0 +1,132 @@
...
+    val adaptedDisconnectTick: Option[Long] =
+      if ev.departureTick > forecastEnd then {
+        // departure is after forecast horizon, we thus ignore it
+        None
+      } else {
+        // we have to provide options for at least one step
+        // thus we can depart at this tick earliest
+        val earliestEnd = currentTick + forecastResolution
+
+        if ev.departureTick < earliestEnd then Some(earliestEnd)
```
staudtMarius, line 75: Shouldn't this be `None`, or `ev.departureTick`, since we can't charge the afterwards?

src/test/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModelSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+      val cases = Table(
+        ("ticksUntilDeparture", "expectedTicksUntilDisconnect"),
+        /* departure before first step */
+        (1, Some(1800)),
+        (900, Some(1800)),
+        (1800, Some(1800)),
+
+        /* departure close to a step */
+        (2699, Some(1800)),
+        (2700, Some(3600)),
```
staudtMarius, line 65: Does it makes sense to round up here?

staudtMarius (changes requested):
I have two small questions.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
sebastian-peter: This case here happens if the EV stays for less than one step (so we would only have the energy limits for the current tick). Then, we wouldn't be able to (dis-)charge at all. We need at least energy limits at two ticks to enable (dis-)charging.

src/test/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModelSpec.scala
sebastian-peter: You mean only round up, never down? My rational was to get the closest tick to actual departure time, which means we round up and down. If there's a good reason to only round up, I'd be open to that too.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
staudtMarius: Will the (dis-)charging be stopped after the ev is leaving?

src/test/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModelSpec.scala
staudtMarius: My thought was to only round down, but if there is no problem, because the ev is only charged until its departure, then we can leave it this way.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
sebastian-peter: Yes, that's the point of calculating the disconnect tick. After that tick, the EV battery is considered disconnected by optimization and can no longer be charged or discharged.

src/test/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModelSpec.scala
sebastian-peter: I mean, there would be a way to argue either way... Whenever we round down, we effectively give the EV less time to charge than it has in reality. When we round up, we extend the time it is connected, and also stretch out the duration of the last time step (so calculated set points might not fit reality 100%).
Thinking about it this way, my intuition tells me that indeed instances where we round up and instances where we round down would in the end "cancel each other out". But, I wouldn't spend my time proving this tbh.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
```diff
@@ -0,0 +1,132 @@
...
+        }
+      }
+
+    // we want to have at least the lowest SOC amount in storage
+    // before disconnecting
+    val disconnectingLimits = adaptedDisconnectTick
+      .map { disconnectTick =>
+
+        // energy to charge until departure
+        val regularLowerLimit =
```
danielfeismann, line 93: This is not a power, it is Energy. So maybe 'energyToLowerLimit'?

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
```diff
@@ -0,0 +1,132 @@
...
+        val requiredPower = regularLowerLimit / timeToDeparture
+
+        // since reaching the lower limit needs to be feasible,
+        // we have to adapt the limit if the regular limit is
+        // not reachable with max power
+        val adaptedLowerLimit =
+          if requiredPower > maxPower then
+            // we cannot reach regular lower limit,
+            // thus we do the most we can
+            maxPower * timeToDeparture
```
danielfeismann: Should be maxPower, right?

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
```diff
@@ -0,0 +1,132 @@
...
+          ev.eStorage * model.departureTargetSoc - ev.storedEnergy
+        val timeToDeparture = Seconds(disconnectTick - currentTick)
+
+        // required power to reach the regular lower limit
+        val requiredPower = regularLowerLimit / timeToDeparture
+
+        // since reaching the lower limit needs to be feasible,
+        // we have to adapt the limit if the regular limit is
+        // not reachable with max power
+        val adaptedLowerLimit =
```
danielfeismann, line 103: energyWithAdaptedPowerLimit?

danielfeismann (commented):
Just a comment from my side on variable naming.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
sebastian-peter: I don't understand yet why this code makes you think it's a power?

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
sebastian-peter: We adapt the energy limit though, not the power limit? I'm a bit confused

danielfeismann (approved):
LGTM

staudtMarius (approved):

PR #1695, opened 2026-02-12
Migrate Gradle from Groovy to Kotlin

### LINKED ISSUE #1030: Migrate gradle from Groovy to Kotlin

danielfeismann:
resolves #1030 

second attempt, ~~mostly done by AI.~~ That was the aim. It required much manual work...

Following gradle tasks can not migrated, mostly since extension not being accessible in the Kotlin DSL script context. 
- spotless
- documentaiton
- sonarqube
- scoverage
- deploy
- mavenCentralPublish
- semVer

However groovy and kotlin can co-exist. Please have a look if this is an improvement or else, simply close this one. Wanted to try if AI can support in these kind of things.


sebastian-peter:
Per our discussion, let's go one of two ways with this:
- Either it is possible to completely replcae groovy with kotlin, then someone should complete this
- Or it is not possible, then we should close this issue/PR, because we don't want to have an additional language in simona

PR #1714, opened 2026-02-25
Fixing PV log spamming in the night

### LINKED ISSUE #984: Reduce log spamming during night time due to no solar irradiance
When it is night and no solar irradiance is available we spam the log with warnings:

```
15:26:51.529 INFO  e.i.s.e.l.RuntimeEventListener$ - ******* Simulation until 2023-02-09T22:45:00Z completed. (duration: 0h : 0m : 0s : 15ms , memory: 3.7 GB) ******
15:26:51.542 WARN  e.i.s.s.weather.WeatherSourceWrapper - Direct solar irradiance not available at POINT (7.65 48.3).
15:26:51.542 WARN  e.i.s.s.weather.WeatherSourceWrapper - Direct solar irradiance not available at POINT (7.65 48.05).
15:26:51.542 WARN  e.i.s.s.weather.WeatherSourceWrapper - Direct solar irradiance not available at POINT (7.9 48.3).
15:26:51.542 WARN  e.i.s.s.weather.WeatherSourceWrapper - Direct solar irradiance not available at POINT (7.9 48.05).
15:26:51.545 WARN  e.i.s.s.weather.WeatherSourceWrapper - Direct solar irradiance not available at POINT (7.65 48.3).
```

The warning was intended to signify that there is actually no weather in the database but not that its magnitude is 0. We should differentiate those cases.

sebastian-peter:
Closes #984

Another try from my side. In this solution, I replaced missing values (`null`) with `Double.Nan`, not `0d`. This way, missing and zero values cannot be confused anymore.
I've also taken the liberty to rename `EMPTY_WEATHER_DATA` to `ZERO_WEATHER_DATA`, which makes it a bit clearer what is meant at first glance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `CHANGELOG` entry for #1607 [#1609](https://github.com/ie3-institute/simona/issues/1609)
 - Fixed HP model returning current tick as the next activation [#1622](https://github.com/ie3-institute/simona/issues/1622)
 - Disallowed EMs as inferiors of `PrioritizedFlexStrat` [#1712](https://github.com/ie3-institute/simona/issues/1712)
+- Fixed log spamming during night time due to no solar irradiance [#984](https://github.com/ie3-institute/simona/issues/984)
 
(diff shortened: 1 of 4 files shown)


staudtMarius (approved):
LGTM.

PR #1731, opened 2026-03-13
Move handling of disaggregated flex options to `EmAgent`

### LINKED ISSUE #1633: Handle disaggregated flex options in em agent
Currently, the handling of disaggregated flex options is done by the em service core. This leads to an increased complexity in the service. Since the disaggregated information is present in the em agent, the handling should be moved there.

staudtMarius:
Resolves #1633


src/main/scala/edu/ie3/simona/agent/participant/ParticipantAgent.scala
sebastian-peter, line 1: Why not make the "disaggregated" (maybe "individual" would be a better term?) flex options a new `FlexType`? The new boolean flag in `FlexActivation` only applies to PowerLimitFlexOptions now.

staudtMarius: For an external Simulation, I need to be able to request them in either form during the simulation. Currently, we handle only PowerLimitFlexOptions in the ExtEmService.

PR #1764, opened 2026-04-09
Moving next tick determination to `ParticipantFlexModel`

### LINKED ISSUE #1692: Move next tick determination to `ParticipantFlexModel`
Currently, `def determineOperatingPoint(state: S, setPower: Power ): (OP, OperationChangeIndicator)` determines the next tick (within OperationChangeIndicator).
With the introduction of energy boundaries, next tick needs to be determined in a different way though. Thus, it now becomes apparent that this functionality should belong with the flex model, not with the participant model.

sebastian-peter:
Closes #1692 

Merge first:
- #1763
- #1766


src/main/scala/edu/ie3/simona/model/participant/ParticipantModelShell.scala
staudtMarius, line 274: What is the point of defining a method here? It is only used once.

staudtMarius (changes requested):
Looks good to me. I have just one small question.

src/main/scala/edu/ie3/simona/model/participant/ParticipantModelShell.scala
sebastian-peter: It is passed to `determineOperatingPoint`... Defining it inline wouldn't look too clean, I'd say

staudtMarius: Okay, But you coud make this a `val` instead of `def`.

sebastian-peter: I could make this a `val modelOperatingPoint: () => (OP, OperationChangeIndicator) = { () => ... }`, but I don't know what would be the benefit of that.
I intentionally made this a function so that `determineOperatingPoint` only executes the calculation if we're inside operation interval.

staudtMarius: I see. In that case, we should keep it this way.

staudtMarius (approved):

PR #1766, opened 2026-04-09
Simplifications in `ThermalGrid`

### LINKED ISSUE #1765: Simplifications in `ThermalGrid`
Also separating operating point from threshold calculation in preparation for #1692

sebastian-peter:
Closes #1765

### DIFF
--- a/src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
+++ b/src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
@@ -110,20 +110,12 @@ final case class ThermalGrid(
 
     val updatedDomesticHotWaterStorageState = domesticHotWaterStorage
       .zip(lastState.domesticHotWaterStorageState)
-      .map {
-        case (
-              storage: DomesticHotWaterStorage,
-              waterStorageState: ThermalStorageState,
-            ) =>
-          storage.determineState(
-            tick,
-            waterStorageState,
-            waterStorageQDot,
-          )
-        case _ =>
-          throw new IllegalStateException(
-            "Could not find state of domestic hot water storage."
-          )
+      .map { case (storage, waterStorageState) =>
+        storage.determineState(
+          tick,
+          waterStorageState,
+          waterStorageQDot,
+        )
       }
 
     ThermalGridState(
@@ -280,7 +272,7 @@ final case class ThermalGrid(
     val (qDotHouse, qDotHeatStorage, qDotWaterStorage) =
       strategy(qDot, heatStorage, domesticHotWaterStorage)
 
-    handleCase(state, qDotHouse, qDotHeatStorage, qDotWaterStorage)
+    handleFeedInCase(state, qDotHouse, qDotHeatStorage, qDotWaterStorage)
   }
 
   /** Selects the strategy how to distribute the thermal power (qDot) from the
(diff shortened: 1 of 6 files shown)


src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
```diff
@@ -600,21 +576,15 @@ class ThermalGridWithHouseAndStorageSpec
...
+        thermalGrid.maybeReviseFeedInFromStorage(hpState) match {
+          case Some(thermalGridOperatingPoint) =>
             thermalGridOperatingPoint shouldBe ThermalGridOperatingPoint(
               zeroKW,
               heatStorage.pThermalMax,
               heatStorage.getpThermalMax * -1,
               zeroKW,
             )
-            nextThreshold shouldBe Some(StorageEmpty(5400))
+          case None => fail("Unexpected result")
```
danielfeismann, line 592: Why do you no longer check for the nextThreshold?

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -336,117 +313,29 @@ final case class ThermalGrid(
     * @param qDotDomesticHotWaterStorage
     *   In-feed to the domestic hot water storage.
     * @return
-    *   The operating point of the thermal grid and the next threshold if there
-    *   is one.
+    *   The operating point of the thermal grid.
     */
-  private def handleCase(
+  private def handleFeedInCase(
```
danielfeismann, line 344: Start of ScalaDoc should no longer contain "from ... the thermal grid"

/** Handles the different thermal flows into the thermal grid.

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -150,9 +144,13 @@ final case class ThermalGrid(
...
-    val heatStorageDemand = calculateHeatStorageDemand(thermalGridState)
-    val domesticHotWaterStorageDemand = calculateDomesticStorageDemand(
-      thermalGridState
+    val domesticHotWaterStorageDemand = calculateStorageDemand(
+      domesticHotWaterStorage,
+      thermalGridState.domesticHotWaterStorageState,
+    )
+    val heatStorageDemand = calculateStorageDemand(
+      heatStorage,
+      thermalGridState.heatStorageState,
```
danielfeismann: I think one can discuss this. After a lot of refactoring, I found it very positive to simply provide the entire thermalGridState through every method and only unpack the required detail further down the line. However, I don't stick to that approach. Should we handle the house similarly if we keep the unpacked alternative?

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
```diff
@@ -458,117 +347,81 @@ final case class ThermalGrid(
...
+    domesticHotWaterStorage
+      .zip(
+        state.thermalGridState.domesticHotWaterStorageState
+      )
+      .map { case (_, storageState) =>
+        // Check if storage can handle the demand
+        val domesticHotWaterDemand =
+          state.thermalDemands.domesticWaterDemandOfHouse
+
+        if storageState.storedEnergy < domesticHotWaterDemand.required then
```
danielfeismann, line 443: With our discussion from yesterday in mind, I would assume that also in this case the missing delta between required and storedEnergy would be lost. Since the storage would provide the stored energy and then refill the storage.

danielfeismann (changes requested):
Thanks a lot for working once through this and for your improvements!

src/test/scala/edu/ie3/simona/model/thermal/ThermalGridWithHouseAndStorageSpec.scala
sebastian-peter: Because it is no longer returned here... I added the check back using `thermalGrid.getThreshold`

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
sebastian-peter: I think your link to #1767 belongs here

danielfeismann: True... Thanks!

src/main/scala/edu/ie3/simona/model/thermal/ThermalGrid.scala
sebastian-peter: I added the two methods back, which now rely on the new astracted method though

danielfeismann (approved):

PR #1816, opened 2026-05-26
Adding current SOE to `EnergyBoundariesFlexOptions`

### LINKED ISSUE #1815: Add current SOE to `EnergyBoundariesFlexOptions`
While the idea of only specifying upward and downward energy potential and assuming the current state of energy to be 0 is nice in theory, it has some practical problems.

Most of all, the current SOE is missing from results, which makes plotting a lot more complex than it needs to be.
Also, incidentally, the way it currently works is harder to grasp and makes tests more difficult to understand, as you always have to keep the initial SOE in mind.

sebastian-peter:
Closes #1815

Also:
- Updating PSDM to 9.1.0
- Updating simonaAPI to 0.15.0


src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
```diff
@@ -92,39 +92,40 @@ class EvcsEnergyBoundariesFlexModel(private val model: EvcsModel)
...
       }
       .getOrElse(SortedMap.empty[Long, ClosedInterval[Energy]])
 
     AssetEnergyBoundaries(
+      currentEnergy = ev.storedEnergy,
       energyLimits = SortedMap(
         currentTick -> new ClosedInterval(
-          -ev.storedEnergy,
-          ev.eStorage - ev.storedEnergy,
+          zeroKWh,
```
staudtMarius, line 127: Will this prevent V2G?

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
```diff
@@ -334,13 +298,9 @@ object EnergyBoundariesFlexOptions
...
       AssetEnergyBoundaries(
+        currentEnergy = currentEnergy,
         energyLimits = SortedMap(
-          currentTick -> new ClosedInterval(
-            // the downward energy potential is the currently stored energy
-            -currentEnergy,
-            // the upward energy potential is the energy to be stored until maximum capacity
-            eStorage - currentEnergy,
-          )
+          currentTick -> new ClosedInterval(zeroKWh, eStorage)
```
staudtMarius, line 303: Why did you change this part?

staudtMarius (changes requested):
I have a few questions regarding the energy limits change.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
sebastian-peter: No, not at all. The point of reference is just different. Beforehand, we assumed that the state of energy at the start of the series is 0 kWh. Now, we assume that it is the actual state of energy, let's say 5 kWh. So, now the minimum SOE is simply 0 kWh, while beforehand it would've been -5 kWh.

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
sebastian-peter: Similar explanation to above... Let me know if I should go more into detail.

src/main/scala/edu/ie3/simona/model/participant/evcs/EvcsEnergyBoundariesFlexModel.scala
staudtMarius: Okay. Thanks for the explanation.

staudtMarius (approved):

staudtMarius:
@sebastian-peter You have to take a look at sonarqube.

PR #1851, opened 2026-06-10
Algorithm for tightening energy boundaries

### LINKED ISSUE #1835: Introduce energy boundaries tightening algorithm

sebastian-peter:
Closes #1835


src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
```diff
@@ -307,6 +331,144 @@ object EnergyBoundariesFlexOptions
...
+        ticks: Seq[Long],
+    ): AssetEnergyBoundaries = {
+      // tolerance for adapted boundaries overlapping
+      given tolerance: Energy = WattHours(1e-6)
+
+      val adaptedTicks = boundaries.tickDisconnect
+        .map { tickDisconnect =>
+          // we only determine energy until tickDisconnect
+          // (after that, the asset is unavailable)
+          ticks.takeWhile(_ <= tickDisconnect)
```
staudtMarius, line 357: Is the `tickDisconnect` part of the adapted ticks? Or could it be excluded?

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
```diff
@@ -307,6 +331,144 @@ object EnergyBoundariesFlexOptions
...
+        case Seq(leftTick, rightTick) =>
+          val sampleTime = Seconds(rightTick - leftTick)
+          val leftLimits = limits.getOrElse(leftTick, throwMissing(leftTick))
+          val rightLimits = limits.getOrElse(rightTick, throwMissing(rightTick))
+
+          val newRightLower = rightLimits.getLower.max(
+            leftLimits.getLower + minNetPower * sampleTime
+          )
+          val newRightUpper = rightLimits.getUpper.min(
+            leftLimits.getUpper + maxNetPower * sampleTime
```
staudtMarius, line 399: Is the `+ maxNetPower` correct here?

staudtMarius (changes requested):
Looks good to me. I have just two small remarks.

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
sebastian-peter: It has to be part of the ticks. Up until disconnection, the asset can be used, which means at disconnection tick we have the last optimization step. This still works in conjunction with our paradigm of "a value/an asset is valid/available up until something changes", because from tickDisconnect and beyond, no energy can be used from the asset.

src/main/scala/edu/ie3/simona/ontology/messages/flex/EnergyBoundariesFlexOptions.scala
sebastian-peter: Yes, we add the maximum possible energy to the last (=left) upper energy limit, and if that amount is below the new (=right) upper boundary, we reduce it to that calculated value, because any values above it would not be possible anyway.

PR #1855, opened 2026-06-17
Introducing split power variable model

### LINKED ISSUE #1837: Implement split power variables optimization model for comparison
Implement split power variables model as MILP and relaxed LP. Include tests.

sebastian-peter:
Closes #1837


src/main/scala/edu/ie3/simona/model/em/opt/CommonLossObjectiveFactory.scala
```diff
@@ -210,15 +154,16 @@ object CommonLossObjectiveFactory {
     */
   class LinearizedQuadraticPowerObjectiveFactory(
       segmentCount: Int
-  ) extends CommonLossObjectiveFactory {
+  ) extends CommonLossObjectiveFactory
+      with PowerVariableObjectiveFactory {
```
staudtMarius, line 158: Since `CommonLossObjectiveFactory` is already extending `PowerVariableObjectiveFactory `, there is no need for the `with` statement. Same for the implementation below.

staudtMarius (changes requested):
Looks good to me. I have just one minor remark.

PR #1862, opened 2026-06-25
Throws exception if external simulation is expected but not found

### LINKED ISSUE #1263: Throw Exception if external simulation is expected but not found
In case we expect but do not find a MobilitySimulation within our SIMONA simulation we currently remove the ev and evcs input and run SIMONA without the MobilitySimulator.

The better option would be to throw an exception and give an Error within the log.

pierrepetersmeier:
Resolves #1263


staudtMarius:
I think that it would be simpler to just throw an exception in `ParticipantAgentFactory.filterValidSysParts()`. We are currently logging a warning there, but we could change it to throwing an exception.

staudtMarius:
@pierrepetersmeier Could you also roll back the changes in `ExtSimSetup`?

input/samples/vn_simona/fullGrid/evcs_input.csv
```diff
@@ -1,3 +1 @@
 uuid,charging_points,cos_phi_rated,id,location_type,node,operates_from,operates_until,operator,q_characteristics,type,v_2g_support,controlling_em
-06a14909-366e-4e94-a593-1016e1455b30,4,0.9,test_evcs_1,HOME,5f1c776c-6935-40f7-ba9e-60646e08992b,,,,"cosPhiFixed:{(0.0,1.0)}",ChargingStationType1,true,
```
staudtMarius, line 2: Could you remove the file too?

staudtMarius (changes requested):
Just one small comment from my side.

staudtMarius (approved):
Looks good to me. Thanks for taking care of this issue.

PR #1876, opened 2026-07-07
Improved parsing of runtime parameters via input files

### LINKED ISSUE #1875: Improve parsing of runtime parameters via input files

staudtMarius:
Resolves #1875

### DIFF
--- a/src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
+++ b/src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
@@ -101,9 +101,14 @@ object ConfigUtil {
         identifier: String,
         additionalParameters: java.util.Map[String, String],
     )(implicit tag: ClassTag[T]): T = {
-      val valid =
-        !additionalParameters.isEmpty && additionalParameters.values.asScala
+      val valid = if !additionalParameters.isEmpty then {
+        val keys = additionalParameters.keySet.asScala
+
+        val fieldNames = tag.runtimeClass.getDeclaredFields.map(_.getName).toSet
+
+        keys.exists(fieldNames.contains) && additionalParameters.values.asScala
           .forall(str => !str.isBlank)
+      } else false
 
       if valid then {
         Try {
--- a/src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
+++ b/src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
@@ -6,7 +6,7 @@
 
 package edu.ie3.simona.util
 
-import com.typesafe.config.ConfigFactory
+import com.typesafe.config.{ConfigFactory, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import edu.ie3.datamodel.io.connectors.{
   CouchbaseConnector,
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
```diff
@@ -56,9 +56,22 @@ import scala.util.{Failure, Success, Try, Using}
...
+  )(using writer: ConfigWriter[T]): T = {
+    val validMap = new util.HashMap(map)
+    validMap.entrySet.removeIf(entry => "".equals(entry.getValue))
+
+    val fallbackValues = ConfigSource.fromConfig(
+      writer.to(fallback).asInstanceOf[ConfigObject].toConfig
+    )
+
+    ConfigSource
+      .fromConfig(ConfigFactory.parseMap(map))
```
PhilippSchmelter: `validMap` is built but never used => filter never applied.

```suggestion
      .fromConfig(ConfigFactory.parseMap(validMap))
```

src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
```diff
@@ -56,9 +56,22 @@ import scala.util.{Failure, Success, Try, Using}
...
 
-  extension (map: java.util.Map[String, String]) {
-    private def toSrc: ConfigObjectSource =
-      ConfigSource.fromConfig(ConfigFactory.parseMap(map))
+  private def load[T <: BaseRuntimeConfig: ConfigConvert](
+      map: java.util.Map[String, String],
+      fallback: T,
+  )(using writer: ConfigWriter[T]): T = {
+    val validMap = new util.HashMap(map)
+    validMap.entrySet.removeIf(entry => "".equals(entry.getValue))
```
PhilippSchmelter: Maybe use `isBlank` to also catch whitespace-only values?

```suggestion
    validMap.entrySet.removeIf(entry => entry.getValue.isBlank)
```

src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
PhilippSchmelter: You could also add a test for this. I think this slipped through.

src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
```diff
@@ -101,43 +114,31 @@ object ConfigUtil {
...
-            throw new RuntimeException(
-              s"No config found of type ${tag.runtimeClass.getSimpleName}."
-            )
         }
+
+        individualConfig
+
+      } match {
+        case Success(conf: T) => conf
+        case _                => defaultConfig
```
PhilippSchmelter: Maybe log a warning on Failure? Otherwise parsing errors silently fall back to the default. Same in `EmConfigUtil` above. L97

PhilippSchmelter (changes requested):
Just some small remarks from me :)

src/main/scala/edu/ie3/simona/util/ConfigUtil.scala
staudtMarius: This may result in a lot of log statements. Therefore, I would not log these.

PhilippSchmelter (approved):
LGTM 🚀

PR #1911, opened 2026-09-04
Added check for data in RunSimonaStandaloneIT

### LINKED ISSUE #1491: Check result data in `RunSimonaStandaloneIT`
There is an old todo in the test `RunSimonaStandaloneIT`. This should be addressed.

See: https://github.com/ie3-institute/simona/pull/1414#discussion_r2224614334

staudtMarius:
Resolves #1491


src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -104,15 +217,9 @@ class RunSimonaStandaloneIT
...
-      // todo implement if valid result handling is implemented (see issue #1491)
-      val pvResultFileContent = getFileSource(
-        resultFileHierarchy,
-        classOf[PvResult],
-      ).getLines().toVector
-      pvResultFileContent.size shouldBe 190
-      pvResultFileContent.headOption.map(
-        _.equals("uuid,inputModel,p,q,timestamp")
-      )
+      simonaSetup.actualResults.foreach { case (key, result) =>
```
PhilippSchmelter, line 223: Do we want to assert completeness here?

We only iterate `actualResults`, so missing results go unnoticed. If the
listener received nothing at all, this passes.

Same goes for L301 below

Also if something does mismatch, the failure only shows the two quantities.
Maybe wrap the call in `withClue(s"$key: ")` so we get the uuid/time/class?

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -270,4 +365,95 @@ class RunSimonaStandaloneIT
...
+
+    (actual, expected) match {
+      case (a: NodeResult, e: NodeResult) =>
+        a.getvAng should equalWithTolerance(e.getvAng)
+        a.getvMag should equalWithTolerance(e.getvMag)
+
+      case (a: CongestionResult, e: CongestionResult) =>
+        a.getSubgrid shouldBe e.getSubgrid
+        a.getType shouldBe e.getType
+        a.getValue should equalWithTolerance(a.getValue)
```
PhilippSchmelter: Comparing 'a' against 'a' looks kinda not right to me. Shouldn't this be compared against 'e'?

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -270,4 +365,95 @@ class RunSimonaStandaloneIT
...
+            a: ElectricalEnergyStorageResult,
+            e: ElectricalEnergyStorageResult,
+          ) =>
+        a.getP should equalWithTolerance(e.getP)
+        a.getQ should equalWithTolerance(e.getQ)
+        a.getSoc should equalWithTolerance(e.getSoc)
+
+      case (a: PowerLimitFlexOptionsResult, e: PowerLimitFlexOptionsResult) =>
+        a.getpMin should equalWithTolerance(e.getpMin)
+        e.getpRef should equalWithTolerance(e.getpRef)
```
PhilippSchmelter: Same here:
`e.getpRef` vs `e.getpRef` should the left side be `a.getpRef`?

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
PhilippSchmelter: Also Lines 382, 383

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -6,37 +6,142 @@
...
 
+  private given tolerance: Double = 1e-7
+
+  // small test setup that adds a new listener
+  final class Setup(
+      override val typeSafeConfig: Config,
+      override val simonaConfig: SimonaConfig,
+      override val args: Array[String] = Array.empty[String],
+      override val runtimeEventQueue: Option[LinkedBlockingQueue[RuntimeEvent]],
+      val actualResults: mutable.Map[
```
PhilippSchmelter: Isn't this map written from the actor thread and read from the test thread after
`run`? Maybe change to thread safe map `TrieMap`

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -6,37 +6,142 @@
...
+        msg.results.values.flatten.foreach { res =>
+          actualResults.put((res.getInputModel, res.getTime, res.getClass), res)
+        }
+
+        Behaviors.same
+      }
+
+      val ref = context.spawn(behavior, "listener")
+
+      super.resultServiceProxy(context, listeners :+ ref, simStartTime)
```
PhilippSchmelter, line 96: Probably a dumb question since I don't know the setup well.
The collector only ends up in the proxy's listener list, so `SimonaSim` doesn't know about it. No watch,
no `FlushAndStop`, nobody waits for it on shutdown. Is that on purpose or would
`resultEventListener` be the better option?

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -6,37 +6,142 @@
...
+        : Map[(UUID, ZonedDateTime, Class[? <: ResultEntity]), ResultEntity] = {
+      val sourse = new ResultEntitySource(expectedResultSource)
+      val tmp = mutable.Map
+        .empty[(UUID, ZonedDateTime, Class[? <: ResultEntity]), ResultEntity]
+
+      def add(res: ResultEntity): Unit =
+        tmp.put((res.getInputModel, res.getTime, res.getClass), res)
+
+      sourse.getNodeResults.forEach(add)
+      sourse.getNodeResults.forEach(add)
```
PhilippSchmelter: I think we can remove one of the lines.

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
```diff
@@ -6,37 +6,142 @@
...
   "A simona standalone simulation" must {
 
+    val expectedResultSource = new CsvDataSource(
+      ",",
+      getResourcePath("vn_simona"),
+      new FileNamingStrategy(),
+    )
+    val expectedResults
+        : Map[(UUID, ZonedDateTime, Class[? <: ResultEntity]), ResultEntity] = {
+      val sourse = new ResultEntitySource(expectedResultSource)
```
PhilippSchmelter: Just a little typo. Rename? :D
Be aware of usages

CHANGELOG.md
```diff
@@ -48,8 +48,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Include functionality description of training data generation for neural state estimation [1709](https://github.com/ie3-institute/simona/issues/1709)
 - Implement energy dependent power tolerance for EVCS [#1698](https://github.com/ie3-institute/simona/issues/1698)
 - Added ground temperatures to weather data [#1818](https://github.com/ie3-institute/simona/issues/1818)
+- Added check for data in `RunSimonaStandaloneIT` [#1491](https://github.com/ie3-institute/simona/issues/1491)
 - Include staging step to Maven Central in CI pipeline [#1834](https://github.com/ie3-institute/simona/issues/1834)
 
+
```
PhilippSchmelter, line 54: I think we can remove this line too!

PhilippSchmelter (changes requested):
Okay it took me a while but here's the awaited review! I think most of my findings are minor changes.

src/test/scala/edu/ie3/simona/integration/RunSimonaStandaloneIT.scala
staudtMarius: We can't use the result event listener, because then we have to read in the data from a temporal folder. But I don't think that we will run into an issue here, since the result proxy, we are using, uses FlushAndStop.

The small dummy listener directly writes the message into the map. Therefore, as long as the result proxy is not terminated, we can write the results correctly.

PhilippSchmelter (approved):
Alright, thanks for including my points.
Looks good to me now!

